PR #1, opened 2024-02-25
feat: permit module

Keyrxng:
Related to https://github.com/ubiquity/.github/issues/98

Cloudflare worker based permit2 module


.env.example
```diff
@@ -1 +1,4 @@
-MY_SECRET="MY_SECRET"
+X25519_PRIVATE_KEY=TNjblUQ1U959TKGddmNlyOJQ7RMGaqS0XXH185Ei1Ik
```
pavlovcik, line 1: Usually you leave values empty in the example.

src/generate-erc20-permit-signature.ts
```diff
@@ -0,0 +1,109 @@
...
+import Decimal from "decimal.js";
+import { decryptKeys } from "./utils/keys";
+
+export async function generateErc20PermitSignature({ beneficiary, amount, issueId, userId }: GenerateErc20PermitSignatureParams) {
+  const config = await configGenerator();
+
+  // @ts-expect-error globalThis
+  globalThis.window = undefined;
+  // @ts-expect-error globalThis
+  globalThis.importScripts = undefined;
```
pavlovcik, line 15: Hope this doesn't cause issues.

src/generate-erc721-permit-signature.ts
```diff
@@ -0,0 +1,167 @@
+import { getPayoutConfigByNetworkId } from "./utils/payoutConfigByNetworkId";
+import { BigNumber, ethers, utils } from "ethers";
+import { MaxUint256 } from "@uniswap/permit2-sdk";
+import { keccak256, toUtf8Bytes } from "ethers/lib/utils";
+
+const NFT_MINTER_PRIVATE_KEY = process.env.NFT_MINTER_PRIVATE_KEY ?? "0x59c6995e998f97a5a0044966f0945389dc9e86dae88c7a8412f4603b6b78690d";
+const NFT_CONTRACT_ADDRESS = "0x38a70c040ca5f5439ad52d0e821063b0ec0b52b6"; // "0x6a87f05a74AB2EC25D1Eea0a3Cd24C3A2eCfF3E0";
```
pavlovcik, line 7: What are these default values?

src/generate-erc721-permit-signature.ts
```diff
@@ -0,0 +1,167 @@
...
+    { name: "deadline", type: "uint256" },
+    { name: "keys", type: "bytes32[]" },
+    { name: "nonce", type: "uint256" },
+    { name: "values", type: "string[]" },
+  ],
+};
+
+const keys = ["GITHUB_ORGANIZATION_NAME", "GITHUB_REPOSITORY_NAME", "GITHUB_ISSUE_ID", "GITHUB_USERNAME", "GITHUB_CONTRIBUTION_TYPE"];
+
+interface Erc721PermitTransactionData {
```
pavlovcik, line 124: I know that you handled consolidating the types. You really should be using an intersection type to generate the 721 schema. 

```
Erc20PermitData & Erc721PermitDataExtension
```

src/generate-permits.ts
pavlovcik, line 1: This shouldnt be here.

src/utils/keys.ts
```diff
@@ -0,0 +1,74 @@
+import SODIUM from "libsodium-wrappers-sumo";
+import { scalarMult, box } from "tweetnacl";
+import tweetnaclUtil from "tweetnacl-util";
+// @ts-expect-error no type
+import blake2b from "blake2b";
```
pavlovcik, line 5: Going to need to do research on these new deps

src/utils/keys.ts
```diff
@@ -0,0 +1,74 @@
...
+ *  and we're good to go
+ *
+ * whether this is viable in production is another question but it's a start
+ */
+
+export async function decryptKeys(cipherText: string): Promise<{ privateKey: string; publicKey: string } | { privateKey: null; publicKey: null }> {
+  let _public: null | string = null;
+  let _private: null | string = null;
+  // this is the evmPrivateEncrypted for 0x70997970C51812dc3A010C7d01b50e0d17dc79C8 (anvil acc[1])
+  const X25519_PRIVATE_KEY = "TNjblUQ1U959TKGddmNlyOJQ7RMGaqS0XXH185Ei1Ik";
```
pavlovcik, line 32: Seems wrong to hard code and post in the open source repo.

wrangler.toml
pavlovcik, line 1: I suppose we can keep this in here to allow for optional Worker deployment. This should really all be invoked from action.yml in the root though, using tsx as the runtime.

pavlovcik (changes requested):

pavlovcik:
I can make an issue soon so you can get credit for this submission.

.env.example
Keyrxng: I was trying to make it easy for testing/debugging and it's the anvil account so security isn't a concern but I'll update this

src/generate-erc721-permit-signature.ts
Keyrxng: the anvil instance of the deployed contract and anvil private key, I'll switch to env

src/utils/keys.ts
Keyrxng: it's the anvil account but I'll switch everything to .env

src/generate-erc721-permit-signature.ts
```diff
@@ -2,12 +2,25 @@ import { getPayoutConfigByNetworkId } from "./utils/payoutConfigByNetworkId";
 import { BigNumber, ethers, utils } from "ethers";
 import { MaxUint256 } from "@uniswap/permit2-sdk";
 import { keccak256, toUtf8Bytes } from "ethers/lib/utils";
+import { Erc721PermitSignatureData, GenerateErc721PermitSignatureParams, PermitTransactionData } from "./types/permits";
 
-const NFT_MINTER_PRIVATE_KEY = process.env.NFT_MINTER_PRIVATE_KEY ?? "0x59c6995e998f97a5a0044966f0945389dc9e86dae88c7a8412f4603b6b78690d";
-const NFT_CONTRACT_ADDRESS = "0x38a70c040ca5f5439ad52d0e821063b0ec0b52b6"; // "0x6a87f05a74AB2EC25D1Eea0a3Cd24C3A2eCfF3E0";
+const NFT_MINTER_PRIVATE_KEY = process.env.NFT_MINTER_PRIVATE_KEY;
+const NFT_CONTRACT_ADDRESS = process.env.NFT_CONTRACT_ADDRESS;
```
pavlovcik, line 8: You should immediately check after if it exists. Generally we use helper functions to 1. import all expected `process.env` values, and 2. check if they exist.

src/utils/keys.ts
```diff
@@ -28,8 +28,8 @@ function deriveNonce(epk: Uint8Array, recipientPubKey: Uint8Array) {
 export async function decryptKeys(cipherText: string): Promise<{ privateKey: string; publicKey: string } | { privateKey: null; publicKey: null }> {
   let _public: null | string = null;
   let _private: null | string = null;
-  // this is the evmPrivateEncrypted for 0x70997970C51812dc3A010C7d01b50e0d17dc79C8 (anvil acc[1])
-  const X25519_PRIVATE_KEY = "TNjblUQ1U959TKGddmNlyOJQ7RMGaqS0XXH185Ei1Ik";
+
+  const X25519_PRIVATE_KEY = process.env.X25519_PRIVATE_KEY;
```
pavlovcik, line 32: Here as well.

src/generate-erc721-permit-signature.ts
pavlovcik: Also I just realized but in a normal node runtime you must use dotenv in every file or else this shouldn't work. Did you test your deliverable? If its not tested this is not ready for review.

Keyrxng: - https://github.com/ubiquity/pay.ubq.fi/pull/177#issue-2152854389 

I posted a test video showing it's all functioning however not within a node env, I'm setting that up now first for ``/research`` then ``/permit``. 

My current setup looks like: 

kernel -> pluginDispatch() -> plugin workflow -> runs the action -> posts the comment

> I thought that plugins will use repository dispatch to return data to the kernel, which will call the next plugin in the chain or stop if it's the last one.

My setup requires that no data be returned to the kernel as my plugins would handle the entire execution flow after dispatch, not sure if that's what you intend for it but it will suffice for testing

pavlovcik: > My setup requires that no data be returned to the kernel as my plugins would handle the entire execution flow after dispatch, not sure if that's what you intend for it but it will suffice for testing

Sure for early testing it might be fine. @whilefoo please chime in on how the interface will look like for returning results from plugins. 

---

The plugins must return results to the kernel because the kernel is what reads the config and connects the chain of plugins together. 

Each partner instance can configure their own order of plugins. 

This is only sensible to implement the logic once inside the kernel to dynamically handle mounting multiple plugins in whatever order.

Keyrxng: Where I'm confused is: invoking an action.yml is different from a workflow, action.yml cannot be executed on a dispatch unless called from a workflow which would imply each plugin will require a workflow listening for it's custom event and the kernel would kick off the workflow which would invoke the action.yml otherwise won't the kernel require some kind of robust generic workflow that it can begin that invokes these action.yml's?

The way I have it set up now, the config.yml is parsed, all plugins for the new comment event  are initiated in plugins, it maps over the regex's applying them then firing the plugin (starting the workflow in the plugin's repo which handles sequencing of the other plugins it relies on by the way of the uses keyword) if it matches any.

If the kernel is piping output from one workflow into the next that's going to take ages running through setup 2/3/4 times as opposed one workflow orchestrating the sequence of other plugins/core actions that it relies on. It's still very modular, partner-configurable and would be quicker than waiting for multiple runners to be assigned, built and executed.

Maybe I'm off on the wrong track, idk but I can't wrap my head around it from a plugin dev standpoint. Partners just add the plugin's info into their config.yml obtained from the plugin's README, the dev defines the config and sequence of called actions and if the kernel is using dispatch to kickoff multiple sequencial workflows then it could just call one workflow that encapsulates everything that plugin needs and then the kernel can be hands-off immediately after dispatch

As custom mounting etc is to be avoided without hindering plugin extensibility I thought it made sense that the kernel scans the config for plugins, applies those regexes on the appropriate webhook events then allowing the plugin's workflow to handle everything.

Are their more reasons to send data back to the kernel other than 'piping'?

pavlovcik: It seems that your primarily concern with this architecture is speed. 

As per my original proposal we can easily compile popular recipes into something like a cloudflare worker if speed is a concern. 

For example the recipe for conversation rewards by default would be three separate GitHub actions in my proposal. But because we use it so frequently it might make sense to compile into a single Worker and use that endpoint as the "workflow" (in this case it would make sense to paste in the entire URL to indicate to the bot that it's an external API)

--- 

Beyond this, it makes sense to conform to GitHub's existing Action schema to take advantage of all of the actions that already exist. We can make the bot interoperate seamlessly with all of those existing capabilities. 

For example, with my proposal it would be trivial to map any action to any arbitrary slash command. The possibilities are infinite though. 

It makes our offering more capable without much extra work. 

---

Lastly it technically would allow us to directly connect our plugins into CI if we wanted to. 

---

I don't understand your concern with action.yml. It can be a simple wrapper that just calls an action inside of the workflows folder. It's a non issue.

Keyrxng: Thanks for this I'm sure I'm on the same page now

Convo Rewards modules:
- commentScoring (action.yml)
- permitGeneration (action.yml)
- permitComment (action.yml)

This recipe is invoked via dispatch (starts a workflow in the recipe repo otherwise it would need to kick off one from the kernel/bot repo which I don't expect) with ``uses`` for each module sequenced as per the workflow file (``ubiquibot-config.plugins?.ISSUE_COMMENT_CREATED`` would match this sequence adhering to the plugin config structure you described

Each module is a lego brick that a workflow imports and runs with or without other modules

If a recipe takes an exorbitant amount of time a worker would be built for that recipe

> I don't understand your concern with action.yml. It can be a simple wrapper that just calls an action inside of the workflows folder. It's a non issue.

Wouldn't it be the other way about, the ``/workflows/*.yml`` calls the re-usable ``action.yml``s and then if piping the output from a workflow (which has run the imported actions) you'd access that through the ``actions/runs/${runID}/jobs`` endpoint?

From what I've read and in building what I have I've found out that dispatching a workflow does not return a runID, you'd need to fetch from ``/actions/runs/`` and filter for ``in_progress`` then match a runner to the workflow that was just dispatched (assuming heavy traffic it might be difficult to find the relevant run). Will know for sure later today

pavlovcik: > Wouldn't it be the other way about, the /workflows/*.yml calls the re-usable action.yml

Don't see why that would be the case. Perhaps there is a gap in my knowledge, but from what I understand `action.yml` is the default entry point for any GitHub Action, so we should follow this. 

Also as a side note, I realize that `conversation-rewards` (the first step of the recipe, which includes qualitative and quantitative analysis of comments) would be impossible to include in a Worker. It was actually the reason why we invented the "delegated compute" infrastructure in the first place (being able to run event handlers outside of the bot kernel.) 

It takes a tremendous amount of time to render comments in a virtual DOM (this is to be able to consolidate both markdown and HTML into a single HTML selector for `ubiquibot-config.yml` purposes) as well as to run the Open AI queries for qualitative analysis. From memory it took ~400ms per comment for quantitative analysis - even running locally on an M1 Macbook - with average issues having around 10 comments to process including their pull requests. And then several network requests to OpenAI, which returns its results over a couple of seconds. 

Workers free tier [has a max of 10ms](https://developers.cloudflare.com/workers/platform/limits/#worker-limits) for perspective.

Keyrxng: Seeing whilefoo's recent push I have a better understanding of how things are going to be running

comment created webhook fires, which I assume uses ``workflow_dispatch`` to start the plugin and the plugin is expected to return the output with a ``repo_dispatch`` dispatch step (I think)

``/research`` needs to be able to read from the botConfig for the openai key, and I see that the authToken is being passed as an input (which I think is insecure if that authToken can read from private repos idk) via ``DelegatedComputeInputs`` I haven't fired any workflows this way yet as I don't want to log anything sensitive(unsure if it would). On my build and with this one that's where I'm stuck.

So I was thinking of setting repo secrets (assuming that's how the @ubiquibot/plugin repos will work too) which in the context of the ``/research-plugin`` repo where the action will run it would have that private read access without sending the app install auth token?

Should my research action be expecting ``npx tsx "body" issueNumber sender org``

or should it be expecting to handle the original event and destructuring everything it needs (token, body, etc..)

Keyrxng:
I'm hanging tight until the plugin setup is configured.

Running in a node env has been tested locally using dotenv and I was able to dispatch events from the kernel, have the script run but it bailed when it needed private repo permissions.

pavlovcik:
> I'm hanging tight until the plugin setup is configured.
> 
> Running in a node env has been tested locally using dotenv and I was able to dispatch events from the kernel, have the script run but it bailed when it needed private repo permissions.

@whilefoo

whilefoo:
@Keyrxng if you have any questions on how to integrate the plugin with the kernel let me know

pavlovcik:
> @Keyrxng if you have any questions on how to integrate the plugin with the kernel let me know

I think the wisest way to proceed is to make a plug-in template repo and everybody can just fork it.

whilefoo:
I already started working on assistive pricing. I can already see that every plugin will need the same setup so in the future we will need to make a plugin SDK.



> I'm starting to wonder if it would make sense to make permit generation in an SDK instead of a restful API.

So every plugin would need its own public/private key to decrypt the partner's private keys, and how we would manage and store all encrypted keys? 
Maybe we can encrypt the private key with the public key of the plugin which should be available to the kernel and this can be also used for all kinds of sensitive things not just private keys.
The config would look like:

```yaml
globals:
  encryptedPrivateKey: "wallet private key encrypted with kernel's private key"
...
- uses:
  - plugin: plugin-A
    with:
       evmEncryptedKey: ${{ encrypt(globals.encryptedPrivateKey) }} # encryptedPrivateKey will be encrypted with plugin's public key 
```
Just some quick thoughts, I still need to think more about this

pavlovcik:
> I already started working on assistive pricing. I can already see that every plugin will need the same setup so in the future we will need to make a plugin SDK.
> 
> > I'm starting to wonder if it would make sense to make permit generation in an SDK instead of a restful API.
> 
> So every plugin would need its own public/private key to decrypt the partner's private keys, and how we would manage and store all encrypted keys? Maybe we can encrypt the private key with the public key of the plugin which should be available to the kernel and this can be also used for all kinds of sensitive things not just private keys. The config would look like:
> 
> ```yaml
> globals:
>   encryptedPrivateKey: "wallet private key encrypted with kernel's private key"
> ...
> - uses:
>   - plugin: plugin-A
>     with:
>        evmEncryptedKey: ${{ encrypt(globals.encryptedPrivateKey) }} # encryptedPrivateKey will be encrypted with plugin's public key 
> ```
> 
> Just some quick thoughts, I still need to think more about this

At a high level I think that it would be best to abstract all the headache of encryption away from partners. The plugin developers should have to worry about it. One, possibly bad idea, is that Ubiquity assigns an encryption key per plugin developer (very similar to how companies provide API keys.) 

Then we can have a single UI that allows partners to generate their configs with encrypted EVM private keys per plugin. The webapp can parse the existing config and then insert the `evmPrivateKeyEncrypted` per plugin in the `ubiquibot-config.yml`.

So the flow would be 1. the user signs in with GitHub 2. they select what repo 3. the webapp parses the config, including the "global" `evmPrivateKeyEncrypted` and then there's a preview of the new config with the correct encrypted keys per plugin 4. the user presses a "save" button, and the UbiquiBot updates their `ubiquibot-config.yml`

Keyrxng:
- https://github.com/ubq-testing/bot-ai/issues/28#issuecomment-1985995745

Nice work foo, once you understand what going on it's pretty intuitive

permit module creating the permit then feeding it into a very simple comment

whilefoo:
> @whilefoo is the intention that once the last `uses` executes it's run it'll send back no data to the kernel? When I was getting things up and running yarn install bugged out with a code 500 and dropped the run all together which the kernel should probs know about but without a response as is, that would imply the end of the line of execution or the kernel would be listening out for a response that'll never come

The plugin should return data to the kernel because it doesn't know if it's the last one in the chain. Even if there are no outputs it should still return empty output object so that the kernel knows it finished execution and calls the next plugin if there is one.

Keyrxng:
> > @whilefoo is the intention that once the last `uses` executes it's run it'll send back no data to the kernel? When I was getting things up and running yarn install bugged out with a code 500 and dropped the run all together which the kernel should probs know about but without a response as is, that would imply the end of the line of execution or the kernel would be listening out for a response that'll never come
> 
> The plugin should return data to the kernel because it doesn't know if it's the last one in the chain. Even if there are no outputs it should still return empty output object so that the kernel knows it finished execution and calls the next plugin if there is one.

This is my plugin config foo:
- generate permit and pass that to comment
- generate comment and post
- Should the comment and post initiate the kernel return data or should there be a 3rd plugin added onto the end which does so?

I'm doubting that I have the config setup the way you envision it but at the same time these two "plugins" which I'm testing with idk whether they'll be bundled as one or used as individual building blocks, but the fact that things are working says I'm on the right track anyway. Once I'm sure I'm on track I'll be more confident in throwing together plugins (`/research`, `/review` etc)/porting features into modules/plugins etc

```ts

plugins:
  issue_comment.created:    
      - name: "Permit Generation"
        description: "Generate a payout permit on the fly."
        command: "^\/permit\\s+((0x[a-fA-F0-9]{40})|([a-zA-Z0-9]{4,})|([a-zA-Z0-9]{3,}\\.eth))\\s+([a-zA-Z0-9]+|\\d+)$"
        example: "/wallet <wallet address>"
        uses:
          - plugin: ubq-testing/permit-generation:compute.yml@plugin-testing
            type: "github" 
            with: 
              evmNetworkId: 100
              evmPrivateEncrypted: ...
              
            id: "permitgeneration"
          - plugin: ubq-testing/payout-comment:compute.yml@development
            type: "github"
            with: 
                type: "${{ permitgeneration.output.type }}"
                permit: "${{ permitgeneration.output.permit }}"
                transferDetails: "${{ permitgeneration.output.transferDetails }}"
                owner: "${{ permitgeneration.output.owner }}"
                signature: "${{ permitgeneration.output.signature }}"
                networkId: "${{ permitgeneration.output.networkId }}"
            id: "permitcomment"
        skipBotEvents: true
```

whilefoo:
The config looks good!

> * Should the comment and post initiate the kernel return data or should there be a 3rd plugin added onto the end which does so?

comment plugin should return data (it can be empty) but it will work even if it doesn't

pavlovcik:
- Exclude `type: "github"`. It should be inferred if there is no protocol prefix i.e. `https://` 
- Also I've never seen `:compute.yml` syntax in GitHub Actions, so lets not do that. It always enters at `action.yml`
- No idea what `skipBotEvents: true` is.

Keyrxng:
> accommodates final permit generation interface of Record<Username, string> i.e. { "pavlovcik": "50" }

@pavlovcik is the intention then that the permit module receive only github username & amount or will there be step between converting username to wallet address or should the permit module be capable of receiving either/or. 

Also, if both nft rewards and erc20 payouts are enabled then how do we distinguish? 

```ts
{
  pavlovcik: {
    amount: 50,
    nft: false,
  },
};
```

whilefoo:
> * Also I've never seen `:compute.yml` syntax in GitHub Actions, so lets not do that. It always enters at `action.yml`

The default is compute.yml but you can change it. If by action.yml you refer to the compose action in the root dir, you can't call a composite action directly.

> * No idea what `skipBotEvents: true` is.

It skips events triggered by bots, for example bot posts a comment

pavlovcik:
> you can't call a composite action directly.

Didn't deeply do research here but I'm very certain that if necessary we can call whatever file we need in a repo, including `action.yml` in the root.

> It skips events triggered by bots, for example bot posts a comment

Makes sense to make that default behavior. I can't think of any instance that we want the bot to react to bots. There are direct ways to handle this via custom webhooks. Otherwise it seems prone to infinite loops. GitHub has a similar apprehension to bot loops (its quite difficult to set up infinite loops with GitHub tooling for bot-to-bot invocations.)

pavlovcik:
> > accommodates final permit generation interface of Record<Username, string> i.e. { "pavlovcik": "50" }
> 
> @pavlovcik is the intention then that the permit module receive only github username & amount or will there be step between converting username to wallet address or should the permit module be capable of receiving either/or.

The permit generation module should, at runtime, look up the used id based on the username. Then it should look in its Cloudflare Key Value storage of the GitHub user ID and the associated wallet address. Then it should generate the permit based on the passed in `keys.evmPrivateEncrypted`

> 
> Also, if both nft rewards and erc20 payouts are enabled then how do we distinguish?
> 
> ```ts
> {
>   pavlovcik: {
>     amount: 50,
>     nft: false,
>   },
> };
> ```

Maybe we should have an NFT permit generation module? It seems that the interface would be different. Maybe the plan is to do something more like 

```json
{ 
   "pavlovcik": {
      "amount": "1",
      "token": "0x4e38D89362f7e5db0096CE44ebD021c3962aA9a0"
   } 
}
```

Keyrxng:
> Then it should look in its Cloudflare Key Value storage of the GitHub user ID and the associated wallet address.

And the permit module is being treated as a plugin technically speaking right? 

@whilefoo only the kernel has access to this if I'm not mistaken, right? The plugins aren't fed the KV id so if the kernel is only processing consecutive plugins, a username > id fetch plugin is required that'll need to run on repo secrets access to the up to date KV key?

Or is the permit module going to be a standalone worker with it's own KV storage?

Maybe I'm missing something...

pavlovcik:
I envision that permit module should be isolated as its own plugin yes. 

@Whilefoo knows implementation details the best but I imagine that it would be most secure to only authenticate requests coming from the kernel. But maybe that's tedious to build?

whilefoo:
> @whilefoo only the kernel has access to this if I'm not mistaken, right? The plugins aren't fed the KV id so if the kernel is only processing consecutive plugins, a username > id fetch plugin is required that'll need to run on repo secrets access to the up to date KV key?

That's correct. if `permit-generation` is the only plugin that needs wallet address of a user then it'd make sense that the plugin also handles a command that sets the wallet address and stores it in plugin's own DB. 

But if multiple plugins need the access to wallet addresses then it depends if the want data to public or private. Making it private is hard because you need a way that only plugins can access it and even then what if somebody makes a plugin just to get wallet addresses. 
If we don't care about addresses being private than it can be simply stored in Supabase DB and made accessible by other plugins via [RLS](https://supabase.com/docs/guides/auth/row-level-security).

> Didn't deeply do research here but I'm very certain that if necessary we can call whatever file we need in a repo, including `action.yml` in the root.

You can't really do that because you are calling a workflow and `action.yml` is a composite action not a workflow, only a workflow can execute an action (https://docs.github.com/en/actions/creating-actions/creating-a-composite-action#testing-out-your-action-in-a-workflow).

pavlovcik:
I think we shouldn't aim to support plugin data sharing yet. Let's build it when we actually need to.

whilefoo:
Is permit generation gonna be a Cloudflare worker or a plugin? @pavlovcik 
@Keyrxng what's left to be done?

Keyrxng:
> Is permit generation gonna be a Cloudflare worker or a plugin? pavlovcik Keyrxng what's left to be done?

Depends on whether or not it's going to be a worker with it's own KV or have it's own supabase DB, I've seen pav say each module should have it's own KV which implies either all modules are independent workers or the kernel has access to the various module KV's and I've seen you say that all plugins should have it's own supabase DB.

I started working on this again about an hour ago, seeing that your going with the supabase DB I've done the same.

TODOs
- handle wallet registration via command
- integrate payout multiplier (seen pav comment it somewhere I forget where)
- decide on how the module will differentiate between ERC20 and 721 payouts if both re active in the partner config

I've got the base setup for either a worker or plugin just needs decided which way it should be taken

pavlovcik:
We should use KV because i think it's a lot simpler to set up for developers for maintenance reasons. 

Although to be fair I didn't work on this so I have limited perspective on if Supabase would be superior for some reason for this use case. 

> Is permit generation gonna be a Cloudflare worker or a plugin

I consider everything a "plugin" and they should support both GitHub actions and cloudflare worker runtimes. I think that it's convenient to make a template that automatically allows plugin developers to run off of GitHub actions but we should accommodate Cloudflare Workers as well for "production" plugins. 

I particularly enjoy the transparency of GitHub actions logs and the other related views.

pavlovcik:
> I think we shouldn't aim to support plugin data sharing yet. Let's build it when we actually need to, given that we are over two weeks behind schedule. 

I realized that we will need to share `/start` command data with `/wallet` registration information. 

This is because `start` should fail if a wallet is not registered. Perhaps `start` command can internally use the same KV instance to simplify the data sharing problem for now. 

@whilefoo rfc

Keyrxng:
I took things in a supabase direction after the initial cloudflare worker build but I can pivot back if it's going to be a worker.

> This is because start should fail if a wallet is not registered. 

I've set it up so that if a wallet cannot be found the permit doesn't get generated and a comment is posted telling them to register and have a reviewer re-run the workflow which I intend to throw if no permit gets generated then it'll get passed to the permit comment plugin once completed. Which in theory should be just pausing it until they register if I'm thinking along the right lines


I've set up 3 commands

- `/wallet (register address)`
- `/permit <wallet addr> <token amount>`
- `/permit <nft address> <username>`

I'm supporting `workflow_dispatch`, `pull_request.closed` && merged, `issue_comment.created`. via the slash commands I can work out if it's an NFT or not but on the first two events it's tougher.

- I'm pulling price from the labels, it would be handy if NFT-based payouts had a `NFT` label something to that affect

<hr/>

> However the old multiplier we used to associate with specific contributors I think we should deprecate. We barely used it.

I started working on `daily-streak` and it's the best way I can think of applying a contributor-based payout increase for streaks

> It would be an even better UX if the plugin generated a dedicated wallet private key for the user as well instead of them bringing their own wallet address. Ideally fully on the client side but an restful API might be okay in the short term

I suggested account abstraction a while back and implemented a rough af version when things started to pivot to a broader target demographic which is basically what you are suggesting here I think. E.g contributor registers via `work.ubq` with login via github and AA kicks in creating a wallet which is saved into db based on hunter's email/github login, then `gasFaucet` funds that address instantly/post-pr

pavlovcik:
> I took things in a supabase direction after the initial cloudflare worker build but I can pivot back if it's going to be a worker.
> 

I think we should only use Supabase if KV isn't sufficient for a plugin's needs (although I can't see how this may be the case.) 

To clarify I have a fair amount of experience using Supabase and conceptually understand KV (haven't really used it much though) but it seems that KV should be faster for plugin developers to set up.

For the short term I say we all try working with KV first for plugins. Then when we eventually make our plugin SDK or plugin template, we can by default include KV. 

> > This is because start should fail if a wallet is not registered.
> 
> I've set it up so that if a wallet cannot be found the permit doesn't get generated and a comment is posted telling them to register and have a reviewer re-run the workflow which I intend to throw if no permit gets generated then it'll get passed to the permit comment plugin once completed. Which in theory should be just pausing it until they register if I'm thinking along the right lines

Permit module should be strictly defined with its capabilities in order to make the code small, simple, and maintanable both from a maintenance perspective and security perspective. The only I/O it must have is with the kernel, which means that it shouldn't directly be able to post error comments. This is a lot of unnecessary code bloat. 

> 
> I've set up 3 commands
> 
> * `/wallet (register address)`
> * `/permit <wallet addr> <token amount>`
> * `/permit <nft address> <username>`
> 

Multiple commands seems interesting but out of scope. We can simply toggle the close as completed state to regenerate permits. Again, this adds unnecessary code bloat. 

> I'm supporting `workflow_dispatch`, `pull_request.closed` && merged, `issue_comment.created`. via the slash commands I can work out if it's an NFT or not but on the first two events it's tougher.
> 
> * I'm pulling price from the labels, it would be handy if NFT-based payouts had a `NFT` label something to that affect
> 

I have no experience working with the NFT code so I have little input here. However the information should be coming in explicitly defined in the config. Implicitly determining if its an NFT reward is the wrong approach.

> > However the old multiplier we used to associate with specific contributors I think we should deprecate. We barely used it.
> 
> I started working on `daily-streak` and it's the best way I can think of applying a contributor-based payout increase for streaks

No I'm referring to the contributor's set multiplier that we saved in the database, and would output their task assignment scoring multiplied with this number. We tried it temporarily but stopped using it. 

If you're starting research on the the daily streak thing, thats fine but it should definitely not be in this code. As per the readme, this codebase should only accept input related to total amount of rewards and to who, then output payment permits only. All other complex logic for calculating the total per contributor etc must be handled outside of this plugin in separate plugins. 

> 
> > It would be an even better UX if the plugin generated a dedicated wallet private key for the user as well instead of them bringing their own wallet address. Ideally fully on the client side but an restful API might be okay in the short term
> 
> I suggested account abstraction a while back and implemented a rough af version when things started to pivot to a broader target demographic which is basically what you are suggesting here I think. E.g contributor registers via `work.ubq` with login via github and AA kicks in creating a wallet which is saved into db based on hunter's email/github login, then `gasFaucet` funds that address instantly/post-pr

I read that Coinbase is working on some in-browser solution that's in beta now. We can probably learn from their implementation and do something similar.

Keyrxng:
> (although I can't see how this may be the case.)

- Either all plugins must be standalone workers then, somehow use the api or have the kernel be able to register new KV stores and assign them to the plugins 

> Multiple commands seems interesting but out of scope.

- The only slash command will be `/wallet`

> The only I/O it must have is with the kernel

- Other than registering the wallet then all info required to generate a permit will be parsed from the previous plugin's output

> However the information should be coming in explicitly defined in the config.

- The previous plugin will pass the required info to determine which reward type it is, if removing the `/permit ...` slash commands there's no need to try to guess.

> thats fine but it should definitely not be in this code

- https://github.com/ubq-testing/ubq-daily-streak, but yes without `/permit` then there is no need to try to grab prices etc, every configurable aspect of the permit will be determined by the previous plugin and this module will generate and output the `PermitTransactionData` object for the next plugin to play with

<hr/>

> For the short term I say we all try working with KV first for plugins.

Plugin devs could setup KV and access it through the API with their cloudflare creds then when the plugin gets ubiquity approval and then becomes hosted under `ubiquity` use org creds and create a KV store for each official plugin? The kernel could be updated to have access to all of these/none of them whatever, if the data isn't sensitive then include the KV ID in the plugin config which other plugin devs could hook into?


> However, if Account A wants to allow Account B to access its KV store, here are some approaches that could be considered:

> API Gateway or Custom Endpoint
> Account A can create an API gateway or a set of custom endpoints within their Worker that proxies requests to the KV store. This Worker would authenticate requests from Account B using a pre-shared key, JWTs, or any other agreed-upon authentication method. This approach effectively allows Account B to read from or write to the KV store by making HTTP requests to the Worker, with the Worker ensuring that only authorized requests are processed.


If I'm to revert back to the worker build I had then the flow should be:

- Plugin A sends permit data to Permit Module
- Module makes req to worker receiving back the `PermitTransactionData` object
- Module returns permit to kernel for Plugin C
><
- User calls `/wallet ...`
- Module sends context to worker, saving meta to KV
- Modules leaves outcome comment
- return `{}` to kernel

pavlovcik:
Similar to the Chrome Extension store, or iOS App Store etc each plugin should be independently hosted with their own backends. This makes development for new developers, as well as managing our infrastructure simple. 

> Edit: A better idea is probably just to have a worker for the sole purpose of KV access and keep everything else as-is that way there's no need for they libsodium workarounds

I'm pretty certain that KV works in GitHub Actions. Besides, worst case scenario we just run "miniflare" inside of GitHub Actions. Its a full Ubuntu environment. It should basically not have any limitations on program execution.

PR #2, opened 2024-03-29
Permits

whilefoo:


package.json
```diff
@@ -40,7 +37,7 @@
     "blake2b": "^2.1.4",
     "decimal.js": "^10.4.3",
     "dotenv": "^16.4.4",
-    "ethers": "^5.7.2",
+    "ethers": "^6.11.1",
```
pavlovcik, line 40: This is a breaking change and I recall that it really changes the API I hope you did this for good reason!

src/handlers/generate-payout-permit.ts
```diff
@@ -1,123 +1,35 @@
...
-      logger.error("No token, amount, spender, or networkId found for ERC20 permit");
-      return "Permit not generated: no token, amount, spender, or networkId found for ERC20 permit";
+export async function generatePayoutPermit(context: Context, permitRequests: PermitRequest[]): Promise<Permit[]> {
+  const permits: Permit[] = [];
+
+  for (const permitRequest of permitRequests) {
+    const { type, amount, username, contributionType } = permitRequest;
+
+    let permit: Permit;
+    switch (type) {
```
pavlovcik, line 19: Why switch case for this

package.json
gentlementlegen: `5.x` seems to be marked legacy so eventually we should bump it up but right now that breaks the code, as it is in the latest commit of that branch.

.github/workflows/compute.yml
```diff
@@ -30,7 +30,7 @@ jobs:
           node-version: "20.10.0"
 
       - name: Install dependencies
-        run: yarn
+        run: npm install -g bun && bun install
```
gentlementlegen, line 33: This can directly be used within https://bun.sh/guides/runtime/cicd
However I don't see where `bun` is used?

src/handlers/generate-payout-permit.ts
gentlementlegen: I think it's fine, if eventually we want to have more permits? It's always better to have a switch case than an if 🌲 forest.

pavlovcik:
> I think it's fine, if eventually we want to have more permits? It's always better to have a switch case than an if 🌲 forest.

We don't have any plans for different types of permits @gentlementlegen 

https://github.com/ubiquibot/permit-generation/pull/2#discussion_r1549107872

src/handlers/generate-payout-permit.ts
pavlovcik: We don't have any plans for different types of permits

PR #10, opened 2024-04-08
feat: encode decode permits

### LINKED ISSUE #8: Permit data structure, encode / decode
The website [pay.ubq.fi](https://pay.ubq.fi/) currently decodes permits passed through an encoded base64 URL, in the form of an array of ECR20 / ECR721  permits, with the following structure
https://github.com/ubiquity/pay.ubq.fi/blob/d193e6dbde2ef3fcd001c6940d50cd0d8b7d722d/scripts/typescript/generate-erc20-permit-url.ts#L37
https://github.com/ubiquity/pay.ubq.fi/blob/4a5d60b4c76855bfb9a800bfdbf0f2adb9955c8c/scripts/typescript/generate-erc721-permit-url.ts#L81
```TypeScript
// ECR20
{
    type: "erc20-permit",
    permit: {
      permitted: {
        token: string,
        amount: string,
      },
      nonce: string,
      deadline: string,
    },
    transferDetails: {
      to: string,
      requestedAmount: string,
    },
    owner:  string,
    signature: string,
    networkId: number,
  };

// ECR721
{
    type: "erc721-permit",
    permit: {
      permitted: {
        token: string,
        amount: string,
      },
      nonce: string,
      deadline: string,
    },
    transferDetails: {
      to: string,
      requestedAmount: string,
    },
    owner: string,
    signature: string,
    networkId: number,
    nftMetadata: {
      GITHUB_ORGANIZATION_NAME,
      GITHUB_REPOSITORY_NAME,
      GITHUB_ISSUE_ID,
      GITHUB_USERNAME,
      GITHUB_CONTRIBUTION_TYPE,
    },
    request: {
      beneficiary: string,
      deadline: string,
      keys: [],
      nonce: string,
      values: [GITHUB_ORGANIZATION_NAME, GITHUB_REPOSITORY_NAME, GITHUB_ISSUE_ID, GITHUB_USERNAME, GITHUB_CONTRIBUTION_TYPE],
    },
  };
```

This repo provides a new structure for the data, as such:
https://github.com/ubiquibot/permit-generation/blob/294cefc17b5dcd9aeb117e8b56d2c2820bce1133/src/types/permits.ts#L3
```TypeScript
export interface Permit {
  tokenType: TokenType;
  tokenAddress: string;
  beneficiary: string;
  amount: string;
  nonce: string;
  deadline: string;
  owner: string;
  signature: string;
  networkId: number;
  erc721Request?: {
    keys: string[];
    values: string[];
    metadata: {
      GITHUB_ORGANIZATION_NAME: string;
      GITHUB_REPOSITORY_NAME: string;
      GITHUB_ISSUE_ID: string;
      GITHUB_USERNAME: string;
      GITHUB_CONTRIBUTION_TYPE: string;
    };
  };
}
```

The problem is that now [pay.ubq.fi](https://pay.ubq.fi/) is not capable to read the generated permits properly. I believe that having this library handling encode / decode would provide a unique source of truth avoiding this situation. There would be two possible approaches:
- accommodate the new code to match the old data shape
- change [pay.ubq.fi](https://pay.ubq.fi/) to read the new object data shape

Downside of the second option is that it would make all permits previously issued impossible to claim.

gentlementlegen:
Resolves #8


gentlementlegen:
To be resolved here: https://github.com/ubiquibot/permit-generation/pull/7#discussion_r1555958708

0x4007:
I guess we will migrate to Cloudflare KV in another future task?

tests/generate-erc20-permit.test.ts
```diff
@@ -80,7 +80,7 @@ describe("generateErc20PermitSignature", () => {
     (context.adapters.supabase.wallet.getWalletByUserId as jest.Mock).mockReturnValue(null);
 
     await expect(async () => {
-      await generateErc20PermitSignature(SPENDER, amount, context);
+      await generateErc20PermitSignature(context, SPENDER, amount);
```
0x4007, line 83: Why did you move this back to the first parameter? I recall earlier in the conversation that we agreed optional parameters should be the last?

gentlementlegen: Relates to the conversation [here](https://github.com/ubiquibot/permit-generation/pull/7#discussion_r1555450639)
Parameters were not optional, just overloaded.

gentlementlegen:
What's going on in this PR:

It introduces function helpers to serialize / deserialize permits and transform them into a base64 encoded string. This is useful when we want to pass the permit data around, like for example with `pay.ubq.fi` within the claim inside the URL.

Having it inside this package allows us to have only one source of truth, easily implementable inside any of our project if needed.

Another problem was that the Supabase generated classes were very old and using the old DB schema. It got updated and code / tests have been fixed accordingly.

gentlementlegen:
> I guess we will migrate to Cloudflare KV in another future task?

@0x4007 what part of Cloudflare KV do you refer when talking about this repo?

molecula451 (approved):
wow, clean stuff, great PR it is the gently GENTLEman......

molecula451 (approved):
wow, clean stuff, great PR it is the gently GENTLEman......

knip.ts
```diff
@@ -4,7 +4,7 @@ const config: KnipConfig = {
   entry: ["src/main.ts"],
   project: ["src/**/*.ts"],
   ignore: [],
-  ignoreBinaries: ["publish"],
+  ignoreBinaries: ["publish", "src/adapters/supabase/types/database.ts"],
```
gitcoindev, line 7: Are you sure this needs to be added to ignoreBinaries?

src/handlers/encode-decode.ts
```diff
@@ -0,0 +1,93 @@
...
+              nonce: permit.nonce,
+              values: permit.erc721Request?.values,
+            },
+          };
+        } else {
+          return null;
+        }
+      }
+    })
+    .filter((o) => o) as RewardPermit[];
```
gitcoindev, line 62: This looks clean.

tests/encode-decode.test.ts
```diff
@@ -0,0 +1,43 @@
...
+    const permitErc721: Permit = {
+      beneficiary: "ubiquity2",
+      deadline: "2",
+      networkId: 101,
+      nonce: "0x2",
+      owner: "0x2",
+      signature: "0x2",
+      tokenAddress: "0x2",
+      tokenType: TokenType.ERC721,
+      amount: 101.5,
```
gitcoindev, line 26: Just curious, isn't ERC721 (NFT reward) amount always supposed to equal 1?

gitcoindev (changes requested):
I left a few questions, but overall a good job!

knip.ts
gentlementlegen: Good catch, that was me experimenting with Knip to ignore the auto-generated code from Supabase, that was causing problems. The fix was to add it ignored inside the `.eslintrc`. Will fix

tests/encode-decode.test.ts
gentlementlegen: This one I do not have an answer for. There was no specific check that I have seen inside previous / other projects. If that's the case it would be worth adding checks around this. @0x4007 might know more

0x4007: @whilefoo @rndquu rfc

But yes intuitively I would say 1 is the correct answer.

0x4007:
> > I guess we will migrate to Cloudflare KV in another future task?
> 
> 
> 
> @0x4007 what part of Cloudflare KV do you refer when talking about this repo?

I want plugins "agents" to default to using KV instead of Supabase. 

This is because I want to make the developer experience as simple as possible and I believe that setting up a KV store is less involved than a Supabase database.

tests/encode-decode.test.ts
molecula451: nice catch gitcoindev, yes non-fungible token, in pieces of 1

gentlementlegen: I made the type enforcing "0" | "1" for ERC721 permits to avoid making this mistake again

gentlementlegen:
> > > I guess we will migrate to Cloudflare KV in another future task?
> > 
> > 
> > @0x4007 what part of Cloudflare KV do you refer when talking about this repo?
> 
> I want plugins "agents" to default to using KV instead of Supabase.
> 
> This is because I want to make the developer experience as simple as possible and I believe that setting up a KV store is less involved than a Supabase database.
> 
> We should consider full database storage as exceptions instead of the norm.
> 
> I also think it's worth implementing the changes in this repository (in another pull) because it will be used as an example for future developers.

KV cannot replace what Supabase does, they have completely different purposes. While Supabase can replace KV, opposite is not true. KV only purpose is to do Key-Value pairs, hence its name. While this is totally suitable for simple things like remembering a user wallet, it is not suitable to do updates, joins, queries since it handles only a one to one relationship. And even in the case of wallets here, they contain multiple columns that cannot be represented through KV. To quote the docs

> Workers KV provides low-latency, high-throughput global storage to your [Cloudflare Workers](https://developers.cloudflare.com/workers/) applications. Workers KV is ideal for storing user configuration data, routing data, A/B testing configurations and authentication tokens, and is well suited for read-heavy workloads.

So in this scenario, it would be possible to do it if we drop other columns of the wallets. But that also means we start to scatter data all around.

tests/encode-decode.test.ts
whilefoo: yes, it's always 1 (or 0)

whilefoo:
We also can't run plugins in Github Actions if we use Cloudflare KV (technically you can with a worker proxy but then you need to implement auth...)

src/handlers/generate-erc721-permit.ts
```diff
@@ -74,20 +75,25 @@ export async function generateErc721PermitSignature(
...
     _walletAddress = walletAddress;
-    _userId = await adapters.supabase.user.getUserIdByWallet(_walletAddress);
     if (isIssueEvent(contextOrPermitPayload)) {
       _issueId = contextOrPermitPayload.payload.issue.id.toString();
     } else {
       throw new Error("Issue Id is missing.");
     }
     _organizationName = contextOrPermitPayload.payload.repository.owner.login;
     _repositoryName = contextOrPermitPayload.payload.repository.name;
+    const { data: userData } = await contextOrPermitPayload.octokit.request("GET /user/:id", { id: _userId });
```
whilefoo, line 92: I think octokit has method for fetching user instead of using paths

src/handlers/encode-decode.ts
```diff
@@ -0,0 +1,107 @@
+import { Permit, TokenType } from "../types";
+import { RewardPermit } from "./permit-type";
+
+/**
+ * Returns a base64 encoded string containing all the permit data
+ */
+export function encodePermits(permits: Permit[]) {
```
whilefoo, line 7: why do we need encode and decode? My understanding is that we will use the new permit format everywhere https://github.com/ubiquibot/permit-generation/issues/8#issuecomment-2041920173

src/handlers/generate-erc20-permit.ts
```diff
@@ -1,58 +1,46 @@
...
+  evmNetworkId: number;
+  evmPrivateEncrypted: string;
+  walletAddress: string;
+  issueId: number;
+  logger: Logger;
+}
+
+export async function generateErc20PermitSignature(payload: Payload, userId: number, amount: number): Promise<Permit>;
+export async function generateErc20PermitSignature(context: Context, userId: number, amount: number): Promise<Permit>;
+export async function generateErc20PermitSignature(contextOrPayload: Context | Payload, userId: number, amount: number): Promise<Permit> {
```
whilefoo, line 18: maybe it'd better to use single object? I don't know why but trying to fit two completely different objects into one variable seems strange to me
```js
{
  userId: number;
  amount: number;
  evmNetworkId: number;
  evmPrivateEncrypted: string;
  walletAddress: string;
  issueId: number;
  logger: Logger;
} & {
  context: Context;
  userId: number;
  amount: number;
}
```

whilefoo (changes requested):

src/types/env.ts
```diff
@@ -4,7 +4,7 @@ import "dotenv/config";
 
 export const envSchema = T.Object({
   SUPABASE_URL: T.String(),
-  SUPABASE_KEY: T.String(),
+  SUPABASE_ANON_KEY: T.String(),
```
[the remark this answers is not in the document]
gentlementlegen: Good point. I thought @0x4007 had renamed this inside every repository to have the same name. I will revert this change so it semantically make sense.

src/types/plugin-input.ts
```diff
@@ -13,7 +13,7 @@ export interface PluginInputs<T extends WebhookEventName = SupportedEvents> {
 
 export const permitRequestSchema = T.Object({
   type: T.Union([T.Literal("ERC20"), T.Literal("ERC721")]),
-  username: T.String(),
+  userId: T.Number(),
```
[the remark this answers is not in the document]
gentlementlegen: Understandable. The reason I chose to use the id instead of the name is because it is consistent and immutable unlike the user name. For example it would break for me or @0x4007 that changed their user names recently, unlike our id that remains the same.

src/handlers/generate-erc721-permit.ts
gentlementlegen: It does, but only using the user name and not the user id. It we decide to change it back to the user name I will use octokit's method

src/handlers/encode-decode.ts
gentlementlegen: Correct. The reason I do this in the first place is to avoid breaking all the permits used before the update. To avoid downtime and issues I think that we should maybe make pay.ubq.fi accept old + new permits before rolling the bot update

gentlementlegen: Related issue here: https://github.com/ubiquity/pay.ubq.fi/issues/222

src/handlers/generate-erc20-permit.ts
gentlementlegen: It is strange but inherent to TypeScript way of handling function overloads, as it is not available in JavaScript unlike traditional OOP languages. Problem with using `&` is that I would have to honor both object members, and by using `|` I would have errors like `A is not assignable to B` without a cast which defeats the type checks.
By doing it this way I get full proper auto complete and type check, and is displayed as such in modern IDEs
image

whilefoo (approved):

src/types/plugin-input.ts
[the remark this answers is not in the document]
0x4007: Sorry for being late. My vision is to allow plugins (who can be developed by anybody) to pass in usernames. 

Under the hood, we should look up the ID during this transaction in the kernel (and use the ID in any code we control) 

This is because when the permit generation starts, the entire transaction takes less than a minute. The users involved in the task completion are unlikely to switch their username during the timeframe of the transaction so this will not pose any problems. 

I want plugin developers to be abstracted away from this implantation detail of always having to refer to users with their ID. 

Does that make sense?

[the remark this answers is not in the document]
gentlementlegen: @0x4007 Yep totally makes sense. I was just worried about that scenario where users change their name, if that would break the logic of finding their ID afterwards. But all the previous references to the account get updated by GitHub on every pull request / issue etc. so this problem should not arise. I had reverted back to the `username` string instead of the ID so all good!

0x4007:
> > > > I guess we will migrate to Cloudflare KV in another future task?
> 
> > > 
> 
> > > 
> 
> > > @0x4007 what part of Cloudflare KV do you refer when talking about this repo?
> 
> > 
> 
> > I want plugins "agents" to default to using KV instead of Supabase.
> 
> > 
> 
> > This is because I want to make the developer experience as simple as possible and I believe that setting up a KV store is less involved than a Supabase database.
> 
> > 
> 
> > We should consider full database storage as exceptions instead of the norm.
> 
> > 
> 
> > I also think it's worth implementing the changes in this repository (in another pull) because it will be used as an example for future developers.
> 
> 
> 
> KV cannot replace what Supabase does, they have completely different purposes. While Supabase can replace KV, opposite is not true. KV only purpose is to do Key-Value pairs, hence its name. While this is totally suitable for simple things like remembering a user wallet, it is not suitable to do updates, joins, queries since it handles only a one to one relationship. And even in the case of wallets here, they contain multiple columns that cannot be represented through KV. To quote the docs
> 
> 
> 
> > Workers KV provides low-latency, high-throughput global storage to your [Cloudflare Workers](https://developers.cloudflare.com/workers/) applications. Workers KV is ideal for storing user configuration data, routing data, A/B testing configurations and authentication tokens, and is well suited for read-heavy workloads.
> 
> 
> 
> So in this scenario, it would be possible to do it if we drop other columns of the wallets. But that also means we start to scatter data all around.

My vision is to encode an array (or object) inside of the value for situations like this @gentlementlegen

0x4007:
> We also can't run plugins in Github Actions if we use Cloudflare KV (technically you can with a worker proxy but then you need to implement auth...)

Auth is not relevant to your argument. 

Auth is required for Supabase as well.

0x4007:
> This is a very utopian scenario where we never have to change the shape of the content, and where this data is only specifically needed in one and only spot. What if tomorrow we have that encoded object and wanna add / remove / change a type? KV doesn't have any migration system, we should be writing scripts that update the data. 

Yes this is a valid concern, but at the same time, I really think we should keep the data PER plugin as simple as possible. 

It's hard to anticipate how large of a data structure any possible future plugin developer would use; but in this exceptional case, the plugin developer can switch out the data storage solution to be Supabase (or any other database)

However for most cases I anticipate that a simple KV will suffice. Think of plugins as like, micro services. Each one should only be concerned with storing simple data directly relevant to their simple capability. 

Again, I am trying to make plugin development as convenient and as accessible as possible for new developers so that our ecosystem can grow faster. It is our duty to make plugin development as easy as possible to get started with. 

> And if that data is accessed by any other project, it will subsequently break every one of them. KV also does not handle RLS or rules to access specific data and columns, which we use to control with fine grain of what is accessible or not, which will lead us to have lot of custom code to protect / control access to such data. 

I anticipated that we needed to implement the custom code anyways on the plugin level. However, yes, this implies less safety measures without RLS etc. again, I don't think it's a showstopper if the data is decentralized. If a single plugin's data is compromised, the entire network won't go down. 

We can also consider making wrapper methods in our SDK that could be useful data for many plugins. For example, collect all data related to the completion of the task (include the associated pull) 

And this includes every user, the list of their contributions etc.  

The SDK could also possibly include the generic boilerplate security related features. This would be a conversation for the future but again the objective is to make plugin development as accessible as possible. 

With KV we cannot seed data either, which is very useful for local setup and testing.
> 
> 
> 
> Maybe this can be discussed somewhere else, but I am not sure to understand the need to move away from Supabase? It is extremely easy to setup, robust, free, and helps accessing the data in an ordered fashion. To me KV seems helpful in scenario where you have just a front-end and wanna save some data specific to a user with no relation with any other object, or settings for example. I feel like we are comparing a shovel to a rake, because both of these techs do not aim to achieve the same purpose, and are not really interchangeable

I have intimate experience working with Supabase so I am quite familiar with the product. But to be honest I only really conceptually understand KV, I haven't really worked with it on a serious project etc. 

There is a chance I could be wrong about this, but conceptually it seems faster and simpler to setup if no UI interaction is necessary to set it up (Supabase in my experience does require to sign in to the UI, establish the schema etc)

rndquu:
Overall it seems that we need to:
1. Use a shared supabase instance for all of the "core" plugins (assistive pricing, conversation rewards, wallet registration, etc...). Most of the plugins share the same entities (i.e. DB tables) so keeping everything in a single DB eliminates the hassle of sharing data between plugins.
2. Allow none-core plugin developers to use whatever they want (supabase, cloudflare KV, etc...). 

Another point in favour of a single DB instance for "core" plugins is that many of our web services (`pay.ubq.fi`, `onboard.ubq.fi`, etc...) require data from a shared supabase instance that we use right now. It is very easy to retrieve data from supabase instance using public anon key. Moving to cloudflare KV for each plugin would require implementing some backend service that would collect data from different cloudflare KV instances so that frontend services could read that data whenever they need to.

0x4007:
> Overall it seems that we need to:
> 
> 1. Use a shared supabase instance for all of the "core" plugins (assistive pricing, conversation rewards, wallet registration, etc...). Most of the plugins share the same entities (i.e. DB tables) so keeping everything in a single DB eliminates the hassle of sharing data between plugins.
> 
> 2. Allow none-core plugin developers to use whatever they want (supabase, cloudflare KV, etc...). 
> 
> 
> 
> Another point in favour of a single DB instance for "core" plugins is that many of our web services (`pay.ubq.fi`, `onboard.ubq.fi`, etc...) require data from a shared supabase instance that we use right now. It is very easy to retrieve data from supabase instance using public anon key. Moving to cloudflare KV for each plugin would require implementing some backend service that would collect data from different cloudflare KV instances so that frontend services could read that data whenever they need to.
> 
> 

I think this is a good proposal. So then let's focus on Supabase in the short term, as we are building the key/core/"official" plugins. 

Then later we should make a boilerplate/example plugin for developers to fork which uses a simpler storage solution such as KV. 

I am concerned that new developers will try and study our existing plugins in order to learn how to make their own. If we don't follow our own "best practices" then it may lead to confusion. We will need to be very deliberate about the messaging (example: use this example plugin project as a starting point instead of studying our existing complex ones) for the sake of growing our developer community faster.

gentlementlegen:
@0x4007 I totally see your point and agree that isolating is good, but as @rndquu said there is a lot of data shared within each plugin like wallets, users, GitHub info and so on. And micro-services would generally use an orchestrator (Kubernetes for example) which helps passing data around, or some central point that helps communication between each entity. It is a common scenario, just again not with KV because I really don't think this is meant to be used for this. And actually, most micro-services spin up their own db even with simple data, and would use NoSql to make things lighter.

whilefoo:
> Auth is not relevant to your point of choosing Supabase over KV.
> 
> Auth is required for Supabase as well.
> 
> @whilefoo

Yes Supabase already has auth but if you want to use Cloudflare KV with plugins running on Github Actions then you need a Cloudflare worker to act as a proxy to access the KV and because Cloudflare workers don't have auth, you have to implement your own auth. 
So we would have to create a proxy worker that every plugin developer has to deploy and use in their plugin which defeats the point of using KV because it's easier

0x4007:
> And micro-services would generally use an orchestrator (Kubernetes for example) which helps passing data around, or some central point that helps communication between each entity

I don't have experience with designing microservices. We should follow best practices, and try our best to make it easy for developers to get started with building plugins.

PR #11, opened 2024-04-08
return data to kernel

whilefoo:


src/index.ts
```diff
@@ -60,13 +60,26 @@ async function run() {
...
 
   return "No permit generated";
 }
 
+async function returnDataToKernel(repoToken: string, stateId: string, output: object) {
+  const octokit = new Octokit({ auth: repoToken });
+  await octokit.repos.createDispatchEvent({
+    owner: github.context.repo.owner,
+    repo: github.context.repo.repo,
+    event_type: "return_data_to_ubiquibot_kernel",
```
pavlovcik, line 75: I feel that the name can be more formal. I would need to see the other event types and we can follow some type of schema/pattern.

whilefoo: What do you suggest?

0x4007: Where can we see a list of the other event types from GitHub?

whilefoo: this is a custom event type, I don't think there are any examples from GitHub

PR #13, opened 2024-04-16
feat!: bundle splitting

gentlementlegen:
This aims to split the bundle to avoid importing Node modules in places where they are not required. Now the imports can be as such:

- `@ubiquibot/permit-generation/core`
- `@ubiquibot/permit-generation/types`
- `@ubiquibot/permit-generation/handlers`


gentlementlegen:
Knip complaining for unknown reasons, also happens locally. But there is no unused dependencies, not unused exports in this code.

molecula451 (approved):
looks like the PR adds vanilla update, looks clean

gitcoindev:
> Knip complaining for unknown reasons, also happens locally. But there is no unused dependencies, not unused exports in this code.

Let me check the root cause.

gitcoindev:
> > Knip complaining for unknown reasons, also happens locally. But there is no unused dependencies, not unused exports in this code.
> 
> Let me check the root cause.

It took me almost 3 hours, which was completely ridiculous but I was able to find and fix this.
I had the same observations as you, failures were shown though dependencies were set up and used correctly.
I tried different settings but no luck.

During a debugging session I discovered that Knip parses .gitignore file! 
I found in the .gitignore for this repository three folders used for build artifacts and Knip matched the same folder names in src (source) folder, and excluded those folders for dependency traversal.

The solution was to add a slash so that only build artifacts are not considered :
https://github.com/ubiquibot/permit-generation/pull/13/commits/cbb4b679caee9d4a89a9d10d749cb9c8d7ad1365

Time to sleep now, tomorrow i am back to other issues -)

gitcoindev (approved):

gentlementlegen:
@gitcoindev thanks a lot for you help. This would mean it happens in every other repository potentially? Or just on this one.

gitcoindev:
@gentlementlegen potentially in other repositories as well. The usual place to place build artifacts is `dist` folder, in this case the errors are not shown as other source folders are usually not named 'dist'. It is worth to remember about this behavior though in  cases, when configuration seems correct but knip still detects unused exports.

gentlementlegen:
@gitcoindev Okay I see. The reason why I am not using `dist` is because if I do then we would need to import like `@ubiquibot/permit-generation/dist/...` which looked odd.

gitcoindev:
> @gitcoindev Okay I see. The reason why I am not using `dist` is because if I do then we would need to import like `@ubiquibot/permit-generation/dist/...` which looked odd.

I solved it for `tsup` with a `package.json` exports configuration, see https://github.com/ubiquity/ubiquibot-logger/blob/development/package.json#L23

There is a good article explaining how this works : https://dev.to/tigawanna/building-and-publishing-npm-packages-with-typescript-multiple-entry-points-tailwind-tsup-and-npm-9e7

You hopefully should be able to achieve similar setup with`rollup`, which is used in this repository.

PR #14, opened 2024-04-17
chore: changed esm to cjs

gentlementlegen:
Simply changed esm to cjs because we mostly use cjs in all our projects, and removed `preserveModules` to lighten the generated code.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,7 +11,3 @@ static/dist
 /coverage
 junit.xml
 dist
-
-/core
-/types
-/handlers
--- a/package.json
+++ b/package.json
@@ -65,6 +65,7 @@
     "npm-run-all": "^4.1.5",
     "prettier": "^3.2.5",
     "rollup": "4.14.3",
+    "rollup-plugin-delete": "2.0.0",
     "rollup-plugin-dts-bundle-generator": "1.4.0",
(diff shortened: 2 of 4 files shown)


gitcoindev (approved):
Reviewed and approved from my side.

gentlementlegen:
@gitcoindev Maybe let me try to fix the `dist` output in this PR, will have a look at what you linked and try to implement the same with `rollup`

gitcoindev:
> @gitcoindev Maybe let me try to fix the `dist` output in this PR, will have a look at what you linked and try to implement the same with `rollup`
> 
> Refers to discussion: [#13 (comment)](https://github.com/ubiquibot/permit-generation/pull/13#issuecomment-2060380353)

sure, good luck!

gentlementlegen:
@gitcoindev So it should be solved by giving the paths inside the `package.json` file. However to make it work when imported, other projects should set `"moduleResolution": "Node16"` inside the `tsconfig.json` to properly resolve the paths.

gentlementlegen:
@gitcoindev Correct these options should be viable as well. It is actually used in https://github.com/ubiquity/pay.ubq.fi/pull/224/files#diff-b55cdbef4907b7045f32cc5360d48d262cca5f94062e353089f189f4460039e0L28

Issue #16, opened 2024-06-12
Dollar permits with fees


rndquu:
User stories:
- Partner should be able to set any ERC20 token as a permit reward (we plan to support only USD-pegged coins initially)
- Ubiquity owner should be able to set permit fees in some env variable. Notes:
  - Fees are transferred to Ubiquity treasury, if permits are regenerated then only the difference is sent
  - Use either direct transfer to ensure partner solvency either permits to the treasury
- Ubiquity owner should be able to whitelist some ERC20 tokens so that they don't incur any fees (ex: Dollar token)

This issue touches the following repositories / docs:
- https://github.com/ubiquibot/permit-generation
- https://github.com/ubiquibot/conversation-rewards
- https://github.com/ubiquity/pay.ubq.fi
- partner docs

gitcoindev:
Good proposals. Number 1 seems strange for me, at least from the partner's perspective. I would reject number 2 immediately. Between number 3 and 4, 3 seems the 'proper' but more complex, if requires audit also costly. For number 4, even if treasury pays for gas, I guess the cost would be negligible for L2 permits (or simply treasury would earn 5% - gas fee). Therefore I vote for option '4', and 2nd choice is option '3'.

Keyrxng:
The only other thing I can think of is just a more complex alternative of 4. Automate treasury permits through OZ Defender but it's a long way for a shortcut.

I'd opt for 4, as it's the easiest to implement and like @gitcoindev says the fee is negligible. 3 seems like overkill despite it being my 2nd choice as well.

EresDev:
very well thought out. 

If I am not mistaken, there is a solution in permit2 that combines the pros of your option 2 & 3 and mostly removes the cons of both.  It appears [permitwitnesstransferfrom](https://docs.uniswap.org/contracts/permit2/reference/signature-transfer#single-permitwitnesstransferfrom) allows you to lock the `SignatureTransferDetails` and the bounty hunter can't change the "transfer to" address.  I could be wrong because I haven't been able to fully understand and try it, probably will need some time to play with it to confirm. It also has a batch variation to combine multiple transfers [batch-permitwitnesstransferfrom](https://docs.uniswap.org/contracts/permit2/reference/signature-transfer#batch-permitwitnesstransferfrom) 

[...]

PR #17, opened 2024-06-14
feat: add custom reward token to permit

rndquu:
Related to https://github.com/ubiquibot/permit-generation/issues/16

This PR adds support for custom ERC20 token address in permits. So instead of hardcoding ERC20 reward token address partners will have to set it manually.

How it is expected to work:
1. Partner sets a config for the `conversation-rewards` plugin similar to:
```
- uses:
      - plugin: ubiquibot/conversation-rewards
        type: github
        with:
          evmNetworkId: 100
          evmPrivateEncrypted: "kmpTKq5Wh9r9x5j3U9GqZr3NYnjK2g0HtbzeUBOuLC2y3x8ja_SKBNlB2AZ6LigXHP_HeMitftVUtzmoj8CFfVP9SqjWoL6IPku1hVTWkdTn97g1IxzmjydFxjdcf0wuDW1hvVtoq3Uw5yALABqxcQ"
          erc20RewardToken: "0xe91D153E0b41518A2Ce8Dd3D7944Fa863463a97d" # WXDAI
```
2. On permit generation in the `conversation-rewards` plugin [here](https://github.com/ubiquibot/conversation-rewards/blob/ba853c5d6606c1b694df36ac72050f6d5b0b20fb/src/parser/permit-generation-module.ts#L75-L82) ERC20 reward token address will be read from the config and passed in permit requests

Notice:
- The package version is bumped from `1.2.2` to `1.2.3` although ideally it should be bumped to `2.0.0` since backwards compatibility for the [generatePayoutPermit()](https://github.com/ubiquibot/conversation-rewards/blob/ba853c5d6606c1b694df36ac72050f6d5b0b20fb/src/parser/permit-generation-module.ts#L84) is broken
- I'm not sure if this PR breaks some other plugins
- The next step is to update https://github.com/ubiquibot/conversation-rewards to use the updated [generateErc20PermitSignature()](https://github.com/ubiquibot/permit-generation/compare/development...rndquu:feat/custom-reward-token?expand=1#diff-a85780b2806b3d56cfa46cb3ae9e404ee06a3c1221d0508ef4c8bee91048a784R19) method


src/utils/payoutConfigByNetworkId.ts
```diff
@@ -1,19 +1,13 @@
 // available tokens for payouts
-export const PAYMENT_TOKEN_PER_NETWORK: Record<string, { rpc: string; token: string; decimals: number }> = {
+export const PAYMENT_TOKEN_PER_NETWORK: Record<string, { rpc: string; }> = {
```
Keyrxng, line 2: Since this is just an RPC source we could use `rpc-handler` here to add a bit of redundancy and allows it to scale with other chains. 

I'd think that might be out of scope for this PR but maybe as another issue?

Keyrxng: Just saw the issue you opened

Keyrxng (approved):
tests passing and looks good

also afaik it doesn't break any other plugins

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@ubiquibot/permit-generation",
-  "version": "1.2.2",
+  "version": "1.2.3",
```
gentlementlegen, line 3: This should be taken care of by `release-please` that should actually bump it to `1.3.0` since you committed it with `feat`. To bump it to `2.0.0` you should commit with `feat!`

src/handlers/generate-erc20-permit.ts
```diff
@@ -63,7 +63,7 @@ export async function generateErc20PermitSignature(contextOrPayload: Context | P
     throw new Error(errorMessage);
   }
 
-  const { rpc, token, decimals } = getPayoutConfigByNetworkId(_evmNetworkId);
+  const { rpc } = getPayoutConfigByNetworkId(_evmNetworkId);
```
gentlementlegen, line 66:
```suggestion
  const { rpc } = getPayoutConfigByNetworkId(_evmNetworkId);
```

gentlementlegen:
This should not break any plugin since plugins will use a specific version of the package, meaning they won't be using it with the changes here, unless we update the version within their `package.json`.

package.json
rndquu: So you mean it's better to keep it `1.2.2`, right?

gentlementlegen: Yes I believe so. Should happen by merging these changes inside the main branch.

rndquu: fixed https://github.com/ubiquibot/permit-generation/pull/17/commits/74ac030203f56d341919c359c1ac83360ad18050

gentlementlegen:
@rndquu fine with the content. I just think you might want to amend the commit to change `feat` to `feat!` as this contains breaking changes.

gentlementlegen (approved):

rndquu:
> @rndquu fine with the content. I just think you might want to amend the commit to change `feat` to `feat!` as this contains breaking changes.

As far I understand the changes introduced in this PR touch only the `conversation-rewards` plugin which I'll update to reflect the changes in this PR. Since all of the plugins are under active development I think it's ok to update the version to `1.3.0` instead of `2.0.0`.

whilefoo (approved):

PR #19, opened 2024-06-16
Feat/rpc handler

### LINKED ISSUE #18: Use `@ubiquity-dao/rpc-handler` instead of hardcoded RPC urls
Right now RPC urls used for reading onchain data are hardcoded in [this](https://github.com/ubiquibot/permit-generation/blob/e2633b36678a6e1749eb703dfc87bd294f47018f/src/utils/payoutConfigByNetworkId.ts) file.

What should be done:
- refactor the code to use https://github.com/ubiquity/rpc-handler that fetches the best RPC url for the provided network id
- remove hardcoded RPC URLs

Keyrxng:
Resolves #18 
Requires #17


Keyrxng:
will rebase once 17 is merged

src/adapters/supabase/helpers/user.ts
```diff
@@ -3,6 +3,9 @@ import { SupabaseClient } from "@supabase/supabase-js";
 import { Super } from "./supabase";
 import { Context } from "../../../types/context";
 
+const FAILED_TO_GET_USER = "Failed to get user";
+const SUCCESSFULLY_FETCHED_USER = "Successfully fetched user";
```
0x4007, line 7: Are these being used?

src/utils/get-fastest-provider.ts
```diff
@@ -0,0 +1,27 @@
...
+  };
+
+  return new RPCHandler(config as HandlerConstructorConfig)
+}
+
+export async function getFastestProvider(networkId: number | string): Promise<JsonRpcProvider> {
+  try {
+    const handler = getHandler(networkId)
+    return await handler.getFastestRpcProvider()
+
```
gentlementlegen, line 23:
```suggestion

```

gentlementlegen:
Tests seem to works but some of them do not exit properly, I suspect some import containing async calls that is not awaited, see [here](https://github.com/ubiquibot/permit-generation/pull/19/checks#step:4:62).

src/adapters/supabase/helpers/user.ts
Keyrxng: They are used in this file yes

gentlementlegen:
@Keyrxng A common scenario I encountered is when you import a file (`import 'my-file'`) and within this file some function is directly called such as:
```TypeScript
run().then().catch(e => {})
```
This mean on import, async functions run instantly but are not awaited and run beyond the tear down of Jest, if that helps you.

Keyrxng:
@gentlementlegen I understand what you mean and why that would be the case but there isn't anything like that in either of the two codebases in this case I've just had a look through them.

Unless the `rpc-handler` bundled somehow ends up with something like that in it which I don't think should happen I'm not sure that's the cause in this case but that's a good thing to keep an eye out for ty. 

I think it's directly related to the racing of the RPCs from what I've seen but like I said I've not been able to remove the warning 100% of the time

src/utils/get-fastest-provider.ts
```diff
@@ -0,0 +1,26 @@
+// @ts-expect-error - no types
```
whilefoo, line 1: `rpc-handler` seems to have types?

whilefoo (approved):

src/utils/get-fastest-provider.ts
Keyrxng: The repo does but the package does not for some reason. My test pkg I published has types and it's the same flow that was used to publish but the UBQ pkg doesn't have types in either of the versions on npmjs

https://www.npmjs.com/package/@ubiquity-dao/rpc-handler?activeTab=code

Keyrxng:
> @Keyrxng A common scenario I encountered is when you import a file (`import 'my-file'`) and within this file some function is directly called such as:
> 
> ```ts
> run().then().catch(e => {})
> ```
> 
> This mean on import, async functions run instantly but are not awaited and run beyond the tear down of Jest, if that helps you.

the `rpc-handler` tests with `--detectOpenHandles` shows this twice

```bash
Jest has detected the following 2 open handles potentially keeping Jest from exiting: 

  ●  TLSWRAP

      16423 |       withCredentials
      16424 |     });
    > 16425 |     let response = await fetch(request);
            |                          ^
      16426 |     const isStreamResponse = supportsResponseStream && (responseType ===
 "stream" || responseType === "response");
      16427 |     if (supportsResponseStream && (onDownloadProgress || isStreamResponse)) {
      16428 |       const options = {};
      ```

gentlementlegen:
@Keyrxng Could you then try to:
- hard code these calls and see if that solves the issue
- if it does, might be better to mock these for testing

Keyrxng:
@gentlementlegen I've mocked the handler and return a valid provider instead of racing them. Tests could be refactored to use a local anvil instance like other projects that rely on on-chain interactions if this impl is a problematic approach to tests.

The warning is no longer present so the issue stems from `rpc-handler`. Because all tests pass in that repo I was under the impression that while you'd rather not have the warning it's not that big of a problem but if it is a problem I can spend some more time and find a solution for it.

gentlementlegen:
I think this should be solved, because at the moment you just get a warning that some code is running after Jest tear down, but if that code was actually outputting things using `console.log` Jest would throw an error saying some code is attempting to log after tear down.

Keyrxng:
Resolved in `rpc-handler` and I've linked the updated package locally and it shows the tests passing without the warning present so the last commit will be reverted and after `proxify` is merged things should be fixed.

image

rndquu (approved):

0x4007:
@rndquu Wrote the spec and approved the pull so I think it's an automatic accept 

@gentlementlegen We should consider this for the auto pull merge plugin

0x4007 (approved):
Implicit ok

rndquu:
@Keyrxng Could you fix https://github.com/ubiquibot/permit-generation/actions/runs/9821721152/job/27117704115 ?

Keyrxng:
> @Keyrxng Could you fix https://github.com/ubiquibot/permit-generation/actions/runs/9821721152/job/27117704115 ?

Opened #29

PR #21, opened 2024-06-18
build: development to main

rndquu:
@gentlementlegen As far as I understand when we merge to the `main` branch the package version is bumped and published to npm.


gentlementlegen (approved):
@rndquu Correct, I thought this was the most sensible way to update the package by merging into the production branch. Good to go

PR #35, opened 2024-07-08
chore: fixed Knip and Jest workflows

gentlementlegen:
- removed comments by Jest and added a output on the Action instead
- knip reported runs only on knip Action failure

### DIFF
--- a/.github/workflows/jest-testing.yml
+++ b/.github/workflows/jest-testing.yml
@@ -2,7 +2,6 @@ name: Run Jest testing suite
 on:
   workflow_dispatch:
   pull_request:
-    types: [opened, synchronize]
 
 env:
   NODE_ENV: "test"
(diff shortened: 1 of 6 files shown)


rndquu (approved):

0x4007:
I feel like I see knip fixes once a week. What is the problem? Why don't we make a dynamic solution?

@gitcoindev @gentlementlegen @rndquu rfc

gitcoindev:
@0x4007 I would propose to convert the Knip workflow into a reusable workflow (https://docs.github.com/en/actions/using-workflows/reusing-workflows) . There is a correct Knip setup now but it needs to be propagated into multiple repositories. The differences between Knip workflow configurations between repositories are minor: bun vs yarn and different yarn versions. I can work on this as soon as I finish the plugin screencasts.

0x4007:
I would prefer to just use bun if we need to pick one. Otherwise yarn v1.

rndquu:
> I feel like I see knip fixes once a week. What is the problem? Why don't we make a dynamic solution?
> 
> @gitcoindev @gentlementlegen @rndquu rfc

All of the bot's plugins and our other services are created either from https://github.com/ubiquibot/plugin-template/, either from https://github.com/ubiquity/ts-template. If there are issues in those template repositories (for example with knip workflow) then all repositories created from those templates must be updated accordingly which might be a tedious task.

I agree with @gentlementlegen that making most of the workflows used in templates (like cspell, conventional commits, knip, etc...) [reusable](https://docs.github.com/en/actions/using-workflows/reusing-workflows) should solve an issue of manually updating tons of repositories in case there is an issue in any of the templates.

0x4007:
> I agree with @gentlementlegen that making most of the workflows used in templates (like cspell, conventional commits, knip, etc...) [reusable](https://docs.github.com/en/actions/using-workflows/reusing-workflows) should solve an issue of manually updating tons of repositories in case there is an issue in any of the templates.

@gentlementlegen perhaps you can write the specification for this as I don't have as much context on the problem as you.

Issue #41, opened 2024-07-26
fix package exports


Keyrxng:
The package is not error free when it is used in other codebases because of an issue with main entry of the `package.json`

PR #42, opened 2024-07-26
fix: main entry for error free imports

Keyrxng:
Resolves https://github.com/ubiquibot/permit-generation/issues/41

This does prevent the usage of `../core/` or `../types/` but seeing as the main entry was `core` which includes everything anyway and was the previous default it seems like this is irrelevant?

QA:

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -96,9 +96,9 @@
     "README.md",
     "package.json"
   ],
-  "module": "core/index.esm.js",
-  "main": "core/index.js",
-  "typings": "types/index.d.ts",
+  "module": "dist/core/index.esm.js",
+  "main": "dist/core/index.js",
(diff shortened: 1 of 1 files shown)


gitcoindev (approved):
Setting up imports correctly can be tricky, thank you for fixing this.

gentlementlegen:
Code looks fine, will test with a fresh project and approve if there is not issue.

gentlementlegen (approved):

Keyrxng:
@gentlementlegen both the commit and PR title are correctly named as per our last convo about package publishing, so merging this should update and publish automatically and correctly right?

gentlementlegen:
@Keyrxng It should be after we merge `development` into `main`.

Issue #46, opened 2024-08-29
Duplicated nonces


rndquu:
Nonces for some github issues are duplicated hence reward claiming is broken for those issues.

Check the following reward URLs:
- https://github.com/ubiquibot/user-activity-watcher/issues/14#issuecomment-2317208648
- https://github.com/ubiquibot/command-start-stop/issues/14#issuecomment-2308374805

In the decoded base64 input all nonces (per user) are equal while the expected behavior is that they should be different.

It happens because [here](https://github.com/ubiquibot/permit-generation/blob/b08723fb0ca280162053013b66873da82320f286/src/handlers/generate-erc20-permit.ts#L111) the nonce is derived from user's github id and github issue number while earlier (not sure when the code was updated) it'd been derived from `issue.node_id`.

What should be done:
- make sure nonces are not duplicated both for ERC20 permits and NFT rewards

0x4007:
The property name suggests issue ID not issue number. That would work.

gentlementlegen:
Impacted rewards that I've seen so far:
- [ ] https://github.com/ubiquity/ubiquibot-kernel/issues/95#issuecomment-2319639659
- [ ] https://github.com/ubiquibot/user-activity-watcher/issues/14#issuecomment-2317208648
- [ ] https://github.com/ubiquibot/command-start-stop/issues/6#issuecomment-2299954707

0x4007:
- I was able to claim https://github.com/ubiquity/ubiquibot-kernel/issues/95#issuecomment-2319639659 just now
-

gentlementlegen:
@0x4007 Still not possible to claim for me.

Keyrxng:
> Impacted rewards that I've seen so far:

I added another issue to this list

gentlementlegen:
I will be regenerating the permits listed above and see if they are claimable now.

PR #47, opened 2024-08-30
Feat/46

rndquu:
Related to https://github.com/ubiquibot/permit-generation/issues/46

P.S. Some `ts` types are updated, not sure if it breaks something

### DIFF
--- a/src/handlers/generate-erc20-permit.ts
+++ b/src/handlers/generate-erc20-permit.ts
@@ -9,7 +9,7 @@ export interface Payload {
   evmNetworkId: number;
   evmPrivateEncrypted: string;
   walletAddress: string;
-  issueId: number;
+  issueNodeId: string;
   logger: Logger;
   userId: number;
 }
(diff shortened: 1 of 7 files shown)


0x4007 (commented):
Can you explain why issue id has collisions? This should be unique

Keyrxng:
https://github.com/ubiquibot/conversation-rewards/blob/9853408f7735bebc50eb2679f118ca82b3cac0c0/src/parser/permit-generation-module.ts#L49

The permit-rewards plugin is passing in issue.id so we'll need to update that or permits will be broken

gentlementlegen (approved):
I am good with the changes, but how come this didn't break before?

rndquu:
> Can you explain why issue id has collisions? This should be unique

I remember there was a long github thread at https://github.com/orgs/community/discussions regarding why `node_id` should be used instead of `id` but I couldn't find it. The closest thing I found is [this](https://github.com/orgs/community/discussions/29784) post. 

In short:
1. `node_id` supersedes `id` and is guaranteed to be unique as stated [here](https://docs.github.com/en/graphql/guides/migrating-graphql-global-node-ids): `These IDs will be unique, therefore you can rely on them directly as references.`
2. Github object id's may be changed over time as mentioned [here](https://github.com/orgs/community/discussions/29784)

`node_id` is basically a base64 encoded string of `{type_id}:{type_name}{resource_id}`.

Example:
```
$ echo "MDEwOlJlcG9zaXRvcnkxMzkwOTU1Mzc=" | base64 --decode
010:Repository139095537
``` 

So if multiple repositories (or issues, PRs, users, etc...) have the same resource id their `node_id` will be different because of the `type_name` used.

rndquu:
> https://github.com/ubiquibot/conversation-rewards/blob/9853408f7735bebc50eb2679f118ca82b3cac0c0/src/parser/permit-generation-module.ts#L49
> 
> The conversation-rewards plugin is passing in issue.id so we'll need to update that or permits will be broken

Let me double check

rndquu:
> I am good with the changes, but how come this didn't break before?

It was broken. `conversation-rewards` plugin incorrectly [passes](https://github.com/ubiquibot/conversation-rewards/blob/9853408f7735bebc50eb2679f118ca82b3cac0c0/src/parser/permit-generation-module.ts#L47) issue number instead of issue id hence the collision occurred. 

> https://github.com/ubiquibot/conversation-rewards/blob/9853408f7735bebc50eb2679f118ca82b3cac0c0/src/parser/permit-generation-module.ts#L49
> 
> The conversation-rewards plugin is passing in issue.id so we'll need to update that or permits will be broken

I'll make a PR

Keyrxng (approved):

Issue #45, opened 2024-08-26
Fix permit generation


gentlementlegen:
From [v1.4.x](https://github.com/ubiquibot/permit-generation/tree/v1.4.0), using the package will yield an error:
```
Error: Failed to get token decimals for token: "0x........"
```
There was recently some updates on the RPC package as well, but I don't think this is related. This is quite important to fix because it seems this package is unusable from `v1.4`.

Example run: https://github.com/ubiquibot/conversation-rewards/actions/runs/10554543532/job/29236569211#step:4:125

@Keyrxng rfc

Keyrxng:
@gentlementlegen it seems [1.4.2](https://github.com/ubiquibot/permit-generation/releases/tag/v1.4.2) was only released an hour ago and the workflow is using [1.3.1](https://github.com/gentlementlegen/conversation-rewards/blob/feat/graphql-pr-fetch/package.json)

Unless I am mistaken and it is using 1.4?

gentlementlegen:
@Keyrxng `1.4.2` introduces minor changes (your fix for types) and indeed the main workflow uses `1.3.1`. I was working on another updated on `conversation-rewards` where I tried to upgrade the package, which is when I noticed the issue, and had to downgrade.

Keyrxng:
Are you using the new method that you added to return a single provider or the main entry?

Keyrxng:
Might be the culprit right here

```json
 "@ubiquity-dao/rpc-handler": "^1.1.0"
```

gentlementlegen:
I think that ideally we should use `ethers@v6` everywhere since `v5` will eventually be deprecated and discontinued. But if it is too much trouble we can take care of this later on and downgrade the version to `v5` in the meantime. @0x4007 rfc

Keyrxng:
Migrating to V6 can be quite the task in some repos to be honest but I do agree that we should probably be using it everywhere.

And if that is happening then it makes sense to migrate to `viem` instead of `Ethers` as it's a far more efficient and streamlined package and has became the industry standard if you ask me

rndquu:
The error is gone, issue may be closed as "not planned"

Keyrxng:
another false alarm against the `rpc-handler`, awesome!

PR #50, opened 2024-09-02
build: use rpc-handler v1.3.0

rndquu:
Check [this](https://github.com/rndquu/conversation-rewards/actions/runs/10665863301/job/29560127872) CI run which shows this error:
```
Error: Failed to get token decimals for token: 0xe91D153E0b41518A2Ce8Dd3D7944Fa863463a97d
```

The `permit-generation` module uses `@ubiquity-dao/rpc-handler` `v1.1.0` while the latest one is `v1.3.0`.

I hope the error will be gone after upgrade.

P.S. The test file is modified because networkId 123 actually [exists](https://chainlist.org/chain/123) nowadays hence `networkId=0` is used in order for the expected error to be thrown.
P.P.S. Somehow `ethers@v6` doesn't accept the `JsonRpcProvider` returned by `handler.getFastestRpcProvider` so I had to convert it to `ethers.Provider`. I've double checked that it works as expected.


src/utils/get-fastest-provider.ts
```diff
@@ -16,10 +15,11 @@ function getHandler(networkId: number | string) {
   return new RPCHandler(config as HandlerConstructorConfig);
 }
 
-export async function getFastestProvider(networkId: number | string): Promise<JsonRpcProvider> {
+export async function getFastestProvider(networkId: number | string): Promise<Provider> {
   try {
     const handler = getHandler(networkId);
-    return await handler.getFastestRpcProvider();
+    const provider = await handler.getFastestRpcProvider();
+    return provider as unknown as Provider;
```
Keyrxng, line 22: It's probably safer if you extract the `connection.url` and create a new provider. The issue is that `permit-generation` uses `ethers@v6` and `rpc-handler` uses `ethers@v5` so there are a lot of differences between the two providers.

The downside to that is we lose the benefits of the rpc-handler `Proxy`. We had started discussing it [here](https://github.com/ubiquibot/permit-generation/issues/45#issuecomment-2309567936). `viem` is off the table obviously but in terms of what to do about this repo whether to downgrade or update `rpc-handler` to support both versions, lmk

0x4007: I think we should always use ethers v5 until llms catch up with v6

rndquu: > It's probably safer if you extract the `connection.url` and create a new provider

done

rndquu:
This PR is needed for https://github.com/ubiquity/ubiquibot-kernel/issues/104 so I'm going to merge it.

If there's smth wrong with it then pls let me know, I'll prepare a fix.

gentlementlegen (approved):

0x4007 (commented):
I think it makes a lot more sense to always use the latest of our in house npm modules given the rate of development. 

```json

{
  "dependencies": {
    "package-name": "*"
  }
}

```

@gentlementlegen these pulls, as well as the merges to main for semantic versioning, are just busy work that don't push the needle forward. In fact they inhibit our ability to test the latest changes immediately.  

Our bot config should also target the development branches for every plugin until they are stable. 

I don't appreciate the clutter in my notifications 

image

gentlementlegen:
Develop branches are by nature unstable. I can change all of them to point there but we shall expect much more breaking which is what I tried to avoid. The good point about version Ning was that I know which build broke functionalities. I do agree that it slows down the process in favor of stability. We can switch back to the main branch later on then.

Issue #20, opened 2024-06-17
Whitelist where `keys.evmPrivateEncrypted` can be used


0x4007:
### Stealing Funds

We should figure out a solution for this. It is trivial to: 

1. copy the `keys.evmPrivateEncrypted` as a collaborator in an organization
2. install the bot on your own repository
3. paste in the stolen key
4. generate rewards

We should consider a whitelist system, where an encrypted key can be registered to work with an organization/repo from the same wallet that topped up the balance. This seems to be the most secure and robust way.

In this case, I imagine that we could make a tool where I would authenticate with `ubq.eth` and then I can tell the permit generation module that it has authority to generate permits in specific repositories or organizations with my `keys.evmPrivateEncrypted` found in the organization config. 

It seems a bit involved to build this out, but it is definitely something worth considering in larger organizations. I am curious to hear of proposals for how to solve this business problem more efficiently.

0x4007:
@rndquu rfc

gentlementlegen:
The generation is still located within this repo, and used in several places like `conversation-rewards` for the bot V2 or `comment-incentives` for the current bot. Also this is used in `pay.ubq.fi` to decode and display permit values.

Stealing the key would not be enough, because it is encrypted, You also need the `X25519_PRIVATE_KEY` to decode it. this key is only available through our organization, so to steal it I think you would need access to the secrets. And secrets are not readable so I don't think this can be stolen.

The only way I see right now to generate permits is to trigger the `conversation-rewards` directly by triggering the action through code. Not sure if that case in handled, but I believe it should reject the run.

0x4007:
>  this key is only available through our organization

No its in the kernel runtime secrets, which is a single environment for everywhere the bot is added to via the GitHub App.

rndquu:
Such kind of issues stems from the fact that private key is stored in a public repo (although encrypted). Whitelisting, restricting access, etc... will eventually lead to stolen funds due to some bug or misconfiguration. 

The most robust solution is to store private key on the partner's side but we somehow need to sign permits with partner's private key. All of the permits are saved in a DB right now. We could implement a solution when permits are generated without a signature and later partners simply adds their signatures in an automated way.

Example:
1. Permit without a signature is generated and saved in a DB
2. Partner clones some config repository, sets private key in github secrets. On CRON schedule there will be some script that will sign permits (i.e. adds signature to permits in a DB) for a particular partner using supabase access key with write access to the `permits` table.

[...]

PR #59, opened 2024-09-06
feat: private key formats

rndquu:
Related to https://github.com/ubiquity/ubiquibot-kernel/issues/104

Right now there is an issue when a malicious partner can copy other partner's encrypted private key and use it in its own organization.

This PR introduces 3 private key formats:
1. `PRIVATE_KEY`
2. `PRIVATE_KEY:GITHUB_ORGANIZATION_ID`
3. `PRIVATE_KEY:GITHUB_ORGANIZATION_ID:GITHUB_REPOSITORY_ID`

You may read how they're supposed to be used [here](https://github.com/rndquu/permit-generation/blob/e1f4d39ec68e04b9b7bcfda26742a4e0b42c8016/src/utils/keys.ts#L34-L61).

The next step is to validate organization and repository in the https://github.com/ubiquibot/conversation-rewards plugin if they are allowed to be used in the organization/repository where original issue was called from.

### DIFF
--- a/src/handlers/generate-erc20-permit.ts
+++ b/src/handlers/generate-erc20-permit.ts
@@ -74,11 +74,12 @@ export async function generateErc20PermitSignature(
     throw new Error("Provider is not defined");
   }
 
-  let privateKey = '';
+  let privateKey = null;
   try {
     const privateKeyDecrypted = await decrypt(_evmPrivateEncrypted, String(process.env.X25519_PRIVATE_KEY));
     const privateKeyParsed = parseDecryptedPrivateKey(privateKeyDecrypted);
     privateKey = privateKeyParsed.privateKey;
+    if (!privateKey) throw new Error("Private key is not defined");
   } catch (error) {
     const errorMessage = `Failed to decrypt a private key: ${error}`;
     _logger.error(errorMessage);
--- a/src/handlers/generate-erc20-permit.ts
+++ b/src/handlers/generate-erc20-permit.ts
@@ -2,7 +2,7 @@ import { PERMIT2_ADDRESS, PermitTransferFrom, SignatureTransfer } from "@uniswap
 import { ethers, keccak256, MaxInt256, parseUnits, toUtf8Bytes } from "ethers";
 import { Context, Logger } from "../types/context";
 import { PermitReward, TokenType } from "../types";
-import { decryptKeys } from "../utils";
+import { decrypt, parseDecryptedPrivateKey } from "../utils";
 import { getFastestProvider } from "../utils/get-fastest-provider";
 
 export interface Payload {
@@ -74,10 +74,14 @@ export async function generateErc20PermitSignature(
     throw new Error("Provider is not defined");
   }
 
-  const { privateKey } = await decryptKeys(_evmPrivateEncrypted);
-  if (!privateKey) {
-    const errorMessage = "Private key is not defined";
-    _logger.fatal(errorMessage);
+  let privateKey = '';
+  try {
+    const privateKeyDecrypted = await decrypt(_evmPrivateEncrypted, String(process.env.X25519_PRIVATE_KEY));
+    const privateKeyParsed = parseDecryptedPrivateKey(privateKeyDecrypted);
+    privateKey = privateKeyParsed.privateKey;
+  } catch (error) {
+    const errorMessage = `Failed to decrypt a private key: ${error}`;
+    _logger.error(errorMessage);
     throw new Error(errorMessage);
   }
 
--- a/src/utils/keys.ts
+++ b/src/utils/keys.ts
@@ -61,11 +61,11 @@ export async function getPublicKey(x25519PrivateKey: string): Promise<string> {
  */
 export function parseDecryptedPrivateKey(decryptedPrivateKey: string) {
   let result: {
-    privateKey: string,
+    privateKey: string | null,
     allowedOrganizationId: number | null,
     allowedRepositoryId: number | null,
   } = {
-    privateKey: "",
+    privateKey: null,
     allowedOrganizationId: null,
     allowedRepositoryId: null,
   };
--- a/src/utils/keys.ts
+++ b/src/utils/keys.ts
@@ -1,40 +1,101 @@
 import sodium from "libsodium-wrappers";
-const KEY_PREFIX = "HSK_";
 
-export async function decryptKeys(cipherText: string): Promise<{ privateKey: string; publicKey: string } | { privateKey: null; publicKey: null }> {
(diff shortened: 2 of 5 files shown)


src/handlers/generate-erc20-permit.ts
```diff
@@ -74,10 +74,14 @@ export async function generateErc20PermitSignature(
     throw new Error("Provider is not defined");
   }
 
-  const { privateKey } = await decryptKeys(_evmPrivateEncrypted);
-  if (!privateKey) {
-    const errorMessage = "Private key is not defined";
-    _logger.fatal(errorMessage);
+  let privateKey = '';
```
0x4007, line 77: Empty string seems not the best

src/utils/keys.ts
```diff
@@ -1,40 +1,101 @@
...
+ * @param decryptedPrivateKey Decrypted private key string (in any of the 3 different formats) 
+ * @returns Parsed private key object: private key, organization id and repository id
+ */
+export function parseDecryptedPrivateKey(decryptedPrivateKey: string) {
+  let result: {
+    privateKey: string,
+    allowedOrganizationId: number | null,
+    allowedRepositoryId: number | null,
+  } = {
+    privateKey: "",
```
0x4007, line 68: Here as well

0x4007 (changes requested):

src/handlers/generate-erc20-permit.ts
```diff
@@ -74,10 +74,14 @@ export async function generateErc20PermitSignature(
...
+  try {
+    const privateKeyDecrypted = await decrypt(_evmPrivateEncrypted, String(process.env.X25519_PRIVATE_KEY));
```
Keyrxng, line 79: Unrelated to your PR:

This confuses me a little because this package isn't a worker or workflow but is being used as an npm package and we import the function to create the permit right? But here we are using `process.env.X25519_PRIVATE_KEY`, I am assuming that this works because the env var is defined in `conversation-rewards`?

If `conversation-rewards` is a core plugin then Ubiquity would always be the one generating permits and not partner's which I'm sure is the intention isn't it?

Keyrxng (approved):
I think this might be good to have in the SDK as any plugin which needs access to funds will need it.

It makes it easier for plugin devs to be bad actors, swap out prod workers for malicious ones and logging PKs etc but that's a partner's choice to use unofficial plugins I guess isn't it

src/handlers/generate-erc20-permit.ts
rndquu: fixed https://github.com/ubiquibot/permit-generation/pull/59/commits/4f471f3b91caeb30e9454499d5506fa4a2824de7

src/utils/keys.ts
rndquu: fixed https://github.com/ubiquibot/permit-generation/pull/59/commits/4f471f3b91caeb30e9454499d5506fa4a2824de7

src/handlers/generate-erc20-permit.ts
rndquu: > This confuses me a little because this package isn't a worker or workflow but is being used as an npm package and we import the function to create the permit right?

Yes

> But here we are using `process.env.X25519_PRIVATE_KEY`, I am assuming that this works because the env var is defined in conversation-rewards?

Yes. I suppose that `X25519_PRIVATE_KEY` was left as an "artifact" at the time when `ubiquibot/permit-generation` was meant to be used as a separate plugin.

> If `conversation-rewards` is a core plugin then Ubiquity would always be the one generating permits and not partner's which I'm sure is the intention isn't it?

Ideally partners should generate permits themselves. But since we trade setup complexity for simpler UX (by allowing partners to set their wallet private keys and decrypting them on our side via `X25519_PRIVATE_KEY`) the only way for a partner to generate a permit himself (at least right now) is to:
1. Fork https://github.com/ubiquibot/conversation-rewards
2. Set partner's own `X25519_PRIVATE_KEY`
3. Use a forked version of https://github.com/ubiquibot/conversation-rewards in partner's plugin config

The closest thing we've come so far with the "partners generate permits themselves" is [this](https://github.com/ubiquibot/permit-generation/issues/20#issuecomment-2175340965) issue. In short:
1. Partner sets his wallet private key in github repository secrets in the `ubiquibot-config` repository
2. Kernel calls `ubiquibot-config` as a github action plugin and gets a signed permit

Keyrxng: > when ubiquibot/permit-generation was meant to be used as a separate plugin.

I'm sure the plan is for it still to be used as a plugin eventually but I understand what you are saying.

> the only way for a partner to generate a permit himself (at least right now) is to:

I don't see why we can't pass it as plugin setting when they install the plugin. 

- If it being leaked is a concern then we can append `_token` to it and GitHub will obscure it for us in logs, hacky but effective. 
- If it's org member liability then really isn't that a partner issue as opposed to our problem? But we could address this I think by adding an additional encryption step that comes from us so that without it `X25519_PRIVATE_KEY` alone cannot be used to decrypt `evmPrivateKeyEncrypted`. We could save this salt in `permit-generation` repo secrets and never expose it to contributors, this way partners can pass their encrypted keys around no problem?
- In order to use the encryption UI you should have an official install and we'd prove that before carrying out the encryption, this makes it harder I guess to reverse engineer things. Fits the idea of the 'setup wizard' also as with auth we can write their config files for them etc.

But I see there are plans to turn the config repo into an action and build more utility into that repo which sounds pretty cool although not a lot of reason for it rn if we can pass these as plugin settings vs plugin env.

whilefoo (changes requested):
The decrypt functions should be exported in [src/index.ts](https://github.com/ubiquibot/permit-generation/blob/development/src/index.ts) if we want to use it in `conversation-rewards`

rndquu:
> The decrypt functions should be exported in [src/index.ts](https://github.com/ubiquibot/permit-generation/blob/development/src/index.ts) if we want to use it in `conversation-rewards`

Fixed https://github.com/ubiquibot/permit-generation/pull/59/commits/114a9cfd714bb3afba8d278f1fc45e0348e519ba

src/handlers/generate-erc20-permit.ts
rndquu: There's a lot to cover in the post above. I don't fully understand a complete flow for decrypting partner's `evmPrivateKeyEncrypted` you propose, need an example. Perhaps its better to proceed with this discussion in https://github.com/ubiquibot/permit-generation/issues/20.

whilefoo (approved):

PR #63, opened 2024-09-09
Dev to main

rndquu:


0x4007:
I think let's not do these anymore. Let's just target development instead of main

rndquu:
> I think let's not do these anymore. Let's just target development instead of main

The issue with this approach is that every breaking change or bug breaks the `conversation-rewards` plugin since we're using the latest available version of the `permit-generation` package [here](https://github.com/ubiquibot/conversation-rewards/blob/5bb7895a793242b206a5157a2fdc4d880dd08e4a/package.json#L37).

For example [this](https://github.com/ubiquibot/permit-generation/pull/59) PR introduced breaking changes. With the "target development" approach the `conversation-rewards` plugin would start failing immediately. While with our current approach (`development` + `main` branch for releases) we have time for preparing necessary changes in the `conversation-rewards` plugin keeping backwards compatibility.

0x4007:
> > I think let's not do these anymore. Let's just target development instead of main
> 
> 
> 
> The issue with this approach is that every breaking change or bug breaks the `conversation-rewards` plugin since we're using the latest available version of the `permit-generation` package [here](https://github.com/ubiquibot/conversation-rewards/blob/5bb7895a793242b206a5157a2fdc4d880dd08e4a/package.json#L37).

Yes I understand that. The feedback loop is significantly faster if we do what I'm saying because we run it in various scenarios. 

Besides, in a pinch we could just revert. In reality we would likely be able to push a fix as soon as a problem is noticed. 

This is obviously to speed up development cycles while we can afford to have crashes in production. 

Our team throughput has been low in the last month and we need to do everything we can to pick up the pace.

rndquu:
> > > I think let's not do these anymore. Let's just target development instead of main
> > 
> > 
> > The issue with this approach is that every breaking change or bug breaks the `conversation-rewards` plugin since we're using the latest available version of the `permit-generation` package [here](https://github.com/ubiquibot/conversation-rewards/blob/5bb7895a793242b206a5157a2fdc4d880dd08e4a/package.json#L37).
> 
> Yes I understand that. The feedback loop is significantly faster if we do what I'm saying because we run it in various scenarios.
> 
> Besides, in a pinch we could just revert. In reality we would likely be able to push a fix as soon as a problem is noticed.
> 
> This is obviously to speed up development cycles while we can afford to have crashes in production.
> 
> Our team throughput has been low in the last month and we need to do everything we can to pick up the pace.

https://github.com/ubiquibot/permit-generation/pull/67

PR #66, opened 2024-09-10
Ethers downgrade

### LINKED ISSUE #60: Downgrade to `Ethers@v5`
This repo I think is the only one which is using `Ethers@v6` and it would be best if the package was downgraded and any affected logic be refactored to work with `V5`

CaptainPrinz:
Resolves #60


CaptainPrinz:
@0x4007 Requesting review

0x4007:
Run tests locally and make sure they pass before asking for reviews.

PR #67, opened 2024-09-13
ci: target `development` for releases

rndquu:
Original comment: https://github.com/ubiquibot/permit-generation/pull/63#issuecomment-2347331423

I still think this is not the best idea and I don't see any issues with manually merge `development` into `main` from time to time when we need a new release.

### DIFF
--- a/.github/workflows/publish-package.yml
+++ b/.github/workflows/publish-package.yml
@@ -4,7 +4,7 @@ on:
   workflow_dispatch:
   push:
     branches:
-      - main
+      - development
 
 permissions:
   contents: write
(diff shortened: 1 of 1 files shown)


0x4007:
>  I don't see any issues with manually merge development into main from time to time when we need a new release.

After observing the last month or so, there tends to be between 0.5 - 1 week lag time for doing so. This adds about a week to our feedback loop, slowing down our throughput.

0x4007 (approved):

gentlementlegen (approved):
@rndquu Agreed, I think it's a better process to keep the releasing branch separated from the development one, hopefully it doesn't break too often.

0x4007:
> hopefully it doesn't break too often.

If the breaking doesn't trend downwards/less frequent then we can consider switching back. I anticipate that we should be having less changes over time, and smaller bugs to fix.

PR #69, opened 2024-09-15
feat: add the permit-generator github action

### LINKED ISSUE #68: Convert to "Plugin"
- Make a GitHub Action that allows this repository to receive permit generation requests, and return permits. 
- This should be the only way to generate permits in our system, so that we can enforce capturing value from plugin developers. 
- The `X25519_PRIVATE_KEY` should be in this repository's secrets. 
- Unclear if this is viable, but add footer property so we can promote a message
   - Example: https://github.com/ubiquity/ubiquibot/issues/719#issuecomment-1726227243

### Payload Schema

```
[{ [username]: amount }]
```

On the backend this should convert the username into their user ID for record keeping purposes (they can always change their username)

hhio618:
Resolves #68


.github/workflows/permit-generator.yml
```diff
@@ -0,0 +1,49 @@
...
+        run: |
+          echo "Received input: ${{ github.event.inputs.users_amounts }}"
+          USERS_AMOUNTS=${{ github.event.inputs.users_amounts }}
+          echo "$USERS_AMOUNTS" > users_amounts.json
+
+          cat users_amounts.json | jq -c '.[]' | while read user_amount; do
+            USERNAME=$(echo $user_amount | jq -r 'keys[0]')
+            AMOUNT=$(echo $user_amount | jq -r '.[]')
+
+            # TODO: This will convert the username to it's unique user ID.
```
hhio618, line 38: Is this desired to be in here?

.github/workflows/permit-generator.yml
```diff
@@ -0,0 +1,49 @@
...
+            USER_ID=$(npx tsx scripts/convertUsernameToID.ts $USERNAME)
+
+            #TODO: Generate a permit for the user using the user ID and the amount provided
+            PERMIT=$(npx tsx scripts/generate-permit.ts $USER_ID $AMOUNT)
+
+            # TODO: Implement logic for how to output the permit to the user.
```
hhio618, line 44: Please give me more info about this one.

0x4007: We just need to reply so that we can receive it in the kernel @gentlementlegen @whilefoo how do you think is best?

gentlementlegen: I think that this plugin should not be able to be triggered by anyone or anyone can generate permits of any amount to anywho. Also we can use our permit package that is already covering all of this and have it running inside this plugin. Not sure what benefit we get over just importing and using the package where needed.

0x4007: The original planned architecture is centralized permit generation. 

We need to take a fee. It's part of our business model. 

Permit generation locally was a shortcut done to get plugins working faster.

gentlementlegen: Sure, just we should find of a proper system to specify which wallet is used and how to convey the data safely.

0x4007: > specify which wallet is used

I suppose our centralized database for now.

whilefoo: Well this kinda goes against the rollup feature, because we will be storing rewards/penalties in the database and generating a permit on-demand when the user wants to claim their rewards.

If the vision is that other third-party plugins also generate permits, then are they going to write to our DB with rewards and then fee is taken when the user claims on `pay.ubq.fi` or will they generate permits with `permit-generation` and handle them in their own way?

In the second scenario, this repo can just be an Action plugin that generates a permit and takes a fee like we originally wanted, so you can have a plugin chain with first plugin being the third party plugin that uses their own algorithm to calculate rewards and output them in a format that `permit-generation` plugin wants.

0x4007: Hmm I'm not 100% on implementation details right now but understand that permit generation must be centralized or we can't enforce taking a fee. Perhaps we should discuss specific proposals on how to implement.

whilefoo: > or we can't enforce taking a fee

am I correctly understanding the fee concept as if the reward is 50$, we take for example 1$ and the user gets 49$?

> I think that the kernel should compile all rewards outputs from all plugins per webhook event and then request permits at the end.
> 
> If they are 0 then don't request a permit automatically. Not every webhook event is to generate a reward (like label changes.) handling this automatically will make the developer experience simpler. We just need an SDK method to be able to return requested rewards to the kernel.

The problem with this approach is that the plugin that generated the rewards can't get the permits back from the kernel because it happens after all plugins are already finished executing, so they can't post a comment with permit or any other way they want.

We could combine this approach with the rollup feature so that all plugins tell the kernel the rewards and at the end the kernel stores rewards in the DB and user can claim on `pay.ubq.fi`.

0x4007: > > or we can't enforce taking a fee
> 
> 
> 
> am I correctly understanding the fee concept as if the reward is 50$, we take for example 1$ and the user gets 49$?
> 

Yes

> 
> 
> > I think that the kernel should compile all rewards outputs from all plugins per webhook event and then request permits at the end.
> 
> > 
> 
> > If they are 0 then don't request a permit automatically. Not every webhook event is to generate a reward (like label changes.) handling this automatically will make the developer experience simpler. We just need an SDK method to be able to return requested rewards to the kernel.
> 
> 
> 
> The problem with this approach is that the plugin that generated the rewards can't get the permits back from the kernel because it happens after all plugins are already finished executing, so they can't post a comment with permit or any other way they want.
> 

Situations like this I am proposing that we accommodate by hard coding stuff in the kernel to facilitate. It's not the most elegant approach but it's a viable one. 

> 
> 
> We could combine this approach with the rollup feature so that all plugins tell the kernel the rewards and at the end the kernel stores rewards in the DB and user can claim on `pay.ubq.fi`.
> 
> 
> 
> 

That sounds good to me. I think that roll ups require a bit more trust of the system compared to regular permits. I can see how permits might be appealing to some organizations.

rndquu: As far as I understand we must "centralize" permit generation (i.e. partners shouldn't be able to fork https://github.com/ubiquibot/conversation-rewards plugin and generate permits themselves).

It makes sense to convert the https://github.com/ubiquibot/permit-generation npm package to a cloudflare worker instead of a github action "plugin" because:
1. It fits with the permit rollups (i.e. generate permits on demand). This way `pay.ubq.fi` backend will have to call the `permit-generation` URL to get a signed permit.
2. It fits with the `conversation-rewards` plugin because all we have to do is to refactor it to get a signed permit from a URL instead of an npm package.

I suppose both `conversation-rewards` plugin and `pay.ubq.fi` backend will have to provide some API key in order to prohibit anybody from signing permits.

Regarding plugin developer rewards I always anticipated it working this way:
1. Plugin dev sets a fee in a plugin's `manifest.json`
2. `conversation-rewards` plugin generates an additional fee permit (or saves an additional "reward" in terms of a rolled permites up feature) besides the  original github issue reward permit

@gentlementlegen @whilefoo If it sounds good then I'll updated description for https://github.com/ubiquibot/permit-generation/issues/68.

gentlementlegen: Fine by me, let's just make sure that Cloudflare implementation of `npm:crypto` works properly because a lot of `node` native functions are mocked within Workers, or are experimental. Here is the compatibility list:
https://developers.cloudflare.com/workers/runtime-apis/nodejs/crypto/

whilefoo: > I suppose both `conversation-rewards` plugin and `pay.ubq.fi` backend will have to provide some API key in order to prohibit anybody from signing permits.

That would mean the any third party plugins that want to generate permits will also need the API key and could generate permits whenever they wanted. I think it would be better to restrict plugins so they can only generate permits when the plugin is triggered by an event

rndquu: > That would mean the any third party plugins that want to generate permits

Why would 3rd party plugins want to generate payout permits? Are there any real use cases? It's kind of strange to allow access to partners' wallet private key (by allowing to generate permits) for 3rd parties.

hhio618:
@0x4007 This task looks off  to me. If there's no progress soon, I discontinue my work here on this PR.

.github/workflows/permit-generator.yml
0x4007: I am hoping we can get by without API keys and instead inherit authentication from GitHub or whatever service the request originates from. 

I imagine there must be some way we can verify authentication based on the bot installation and the GitHub repo. 

We could do some research and iterate our authentication as we get further along.

whilefoo: > 4\. The kernel should request these permits to be generated from this api. Because it's our kernel I think that we should be able to verify that it's the kernel somehow

We can basically run it as a Worker plugin, because we can verify it's our kernel because the kernel signs the payload and the plugin verifies it with kernel's public key

.github/workflows/permit-generator.yml
0x4007: We should save user id for audit/accounting reasons. It's confusing if they change their username and that's all we store

0x4007:
Can you post QA to prove it's working

.github/workflows/permit-generator.yml
rndquu: Ok, as far as I understand we keep the `https://github.com/ubiquibot/permit-generation` plugin as a github action.

@whilefoo 

> we can verify it's our kernel because the kernel signs the payload and the plugin verifies it with kernel's public key

A few questions:
1. "kernel signs the payload" is already implemented, correct?
2. `ubiquibot/permit-generation` plugin doesn't verify the caller, correct? I mean this functionality is not implemented right now.
3. Are there any plugins which verify a caller? If so could you share a related code?

whilefoo: > 1. "kernel signs the payload" is already implemented, correct?

Yes



> 2\. `ubiquibot/permit-generation` plugin doesn't verify the caller, correct? I mean this functionality is not implemented right now.

Well technically anybody with workflow permission on the repo can call it but it doesn't verify it's necessarily the kernel

> 3\. Are there any plugins which verify a caller? If so could you share a related code?

Yes, all Worker plugins [verify the caller](https://github.com/ubiquity-os-marketplace/assistive-pricing/blob/9984131a66eb9167dd0f60147a99e41ae382c43b/src/worker.ts#L45)

0x4007:
@whilefoo @rndquu @gentlementlegen I don't recall who worked on the NFT minting part but the tests are failing due to it, and it doesn't seem like any changes related to it were made in this pull.

hhio618:
@0x4007 
QA
------
Workflow dispatch event: https://github.com/hhio618/permit-generation/actions/runs/11167533080/job/31043948081
Report: https://github.com/hhio618/permit-generation/issues/5

0x4007 (approved):
This looks promising, thanks for posting the QA link. 

We can optimize and make an `ncc` build for faster cold start speed, as well as remove the "post comment to issue" action in another pull. 

The purpose of this feature is to centralize permit generation so that we can enforce fees. All plugins should call this as an API with "permit requests."

scripts/github-action-permit-generator.ts
```diff
@@ -0,0 +1,100 @@
...
+      contributionType: "custom",
+      tokenAddress: env.EVM_TOKEN_ADDRESS,
+    }))
+  );
+
+  const config: PermitGenerationSettings = {
+    evmNetworkId: Number(env.EVM_NETWORK_ID),
+    evmPrivateEncrypted: env.EVM_PRIVATE_KEY,
+    permitRequests: permitRequests,
+    runId: `${runId}`,
```
0x4007, line 40:
```suggestion
    runId: runId,
```

scripts/github-action-permit-generator.ts
```diff
@@ -0,0 +1,116 @@
...
+  const userAmounts = JSON.parse(env.USERS_AMOUNTS);
+
+  // Populate the permitRequests from the user_amounts payload
+
+  const permitRequests: PermitRequest[] = userAmounts.flatMap((userObj: { [key: string]: number }) =>
+    Object.entries(userObj).map(([user, amount]) => ({
+      type: "ERC20",
+      username: user,
+      amount: amount,
+      contributionType: "custom",
```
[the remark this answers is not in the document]
whilefoo: Do you mean contributionType? Yes it was for the NFT. If we are generating ERC20 reward we don't need it

permits.json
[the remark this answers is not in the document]
hhio618: Thanks for the heads-up!

0x4007:
> > The purpose of this feature is to centralize permit generation so that we can enforce fees. All plugins should call this as an API with "permit requests."
> 
> 
> 
> But plugins can't call another plugin or do you mean having plugins return permit requests to the kernel and then the kernel calls `permit-generation`?
> 
> 
> 
> I think we discussed that this should be a Worker plugin because the problem Action plugin is that when we implement rollup rewards, the `pay.ubq.fi` backend can't call an Action plugin and get the result back, it can only call a Worker plugin but it needs the kernel's private key to sign the payload.

Got it that makes sense. In this case we can accept this pull and then make a new task for it to be a worker, because this pull achieves the task spec.

scripts/github-action-permit-generator.ts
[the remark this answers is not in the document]
0x4007: Then this line should be removed?

src/types/plugin-input.ts
```diff
@@ -25,6 +25,7 @@ export const permitGenerationSettingsSchema = T.Object({
   evmNetworkId: T.Number(),
   evmPrivateEncrypted: T.String(),
   permitRequests: T.Array(permitRequestSchema),
+  runId: T.Optional(T.Number()),
```
0x4007, line 28: What about this line @whilefoo

whilefoo: I'm not sure why this is needed

hhio618: It is used to generate a unique nonce value; please refer to these lines:
https://github.com/hhio618/permit-generation/blob/b147da4e32f90311b4673311024ccc4b244f377b/src/handlers/generate-erc20-permit.ts#L58
https://github.com/hhio618/permit-generation/blob/b147da4e32f90311b4673311024ccc4b244f377b/src/handlers/generate-erc20-permit.ts#L89

0x4007: @whilefoo

scripts/github-action-permit-generator.ts
```diff
@@ -0,0 +1,116 @@
...
+
+/**
+ * Generates all the permits based on the current github workflow dispatch.
+ */
+export async function generatePermitsFromGithubWorkflowDispatch() {
+  const runId = github.context.runId;
+
+  // These are necessary to ensure the type checks and tests pass.
+  process.env["NFT_MINTER_PRIVATE_KEY"] = "";
+  process.env["NFT_CONTRACT_ADDRESS"] = "";
```
0x4007, line 21: We should fix the type checks in another pull and make these optional probably? @whilefoo rfc

scripts/github-action-permit-generator.ts
0x4007, line 47:
```suggestion
```

src/adapters/supabase/helpers/wallet.ts
```diff
@@ -15,8 +15,11 @@ export class Wallet extends Super {
       throw error;
     }
 
-    console.info("Successfully fetched wallet", { userId, address: data.wallets?.address });
-    return data.wallets?.address;
+    // Check if wallets is an array, if so, return the first element's address
+    const address = Array.isArray(data.wallets) ? data.wallets[0]?.address : data.wallets?.address;
```
0x4007, line 19: There should only be a single wallet associated per user. Why did you check for arrays

src/handlers/generate-erc20-permit.ts
```diff
@@ -54,6 +54,8 @@ export async function generateErc20PermitSignature(
       issueNodeId = contextOrPayload.payload.issue.node_id;
     } else if ("pull_request" in contextOrPayload.payload) {
       issueNodeId = contextOrPayload.payload.pull_request.node_id;
+    } else if (contextOrPayload.config.runId) {
+      issueNodeId = contextOrPayload.config.runId.toString();
```
0x4007, line 58: The node ID should not be the CI run ID? This doesnt make any sense to me

.github/workflows/permit-generator.yml
```diff
@@ -0,0 +1,47 @@
...
+        shell: bash
+        env:
+          X25519_PRIVATE_KEY: ${{ secrets.X25519_PRIVATE_KEY }}
+          EVM_NETWORK_ID: ${{ secrets.EVM_NETWORK_ID }}
+          EVM_PRIVATE_KEY: ${{ secrets.EVM_PRIVATE_KEY }}
+          EVM_TOKEN_ADDRESS: ${{ secrets.EVM_TOKEN_ADDRESS }}
+          SUPABASE_URL: ${{ secrets.SUPABASE_URL}}
+          SUPABASE_KEY: ${{ secrets.SUPABASE_KEY }}
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          USERS_AMOUNTS: ${{ github.event.inputs.users_amounts }}
```
0x4007, line 38: Rename this to payment requests

.github/workflows/permit-generator.yml
```diff
@@ -0,0 +1,47 @@
...
+
+      - name: Report by opening an issue
+        run: |
+          export PERMITS=$(cat permits.txt)
+          curl -X POST \
+            -H "Authorization: token ${{ secrets.GITHUB_TOKEN }}" \
+            -H "Content-Type: application/json" \
+            -d "{\"title\": \"Workflow Dispatch Report\", \"body\": \"$PERMITS\"}" \
+            "https://api.github.com/repos/${{ github.repository }}/issues"
```
0x4007, line 47: I suppose this might make sense to just echo or log in the CI instead

src/types/plugin-input.ts
whilefoo: that makes sense but just using `uuid` to generate a nonce would be sufficient

src/handlers/generate-erc20-permit.ts
whilefoo: Because when you are manually generating a permit with Action workflow dispatch, it's not associated to any issue so you have to use a unique nonce - that could be the run ID or just a uuid

scripts/github-action-permit-generator.ts
whilefoo: yeah this should be done with Typebox

src/handlers/generate-erc20-permit.ts
hhio618: Other ideas could be:

- Pulling it from a workflow input, like `uuid`
- Setting a fixed value that always gives the same nonce each time

src/adapters/supabase/helpers/wallet.ts
0x4007: @whilefoo @rndquu can you concur this is correct?

gentlementlegen: Can't you use `single()` to retrieve only one? The type should be properly deduced.

whilefoo: yes it's normal for that to be an array because it's querying an user with wallets as a relation. if you wanted to get just one record you would need to query the wallet table directly

hhio618: @whilefoo The code is already in the base branch but isn't working without the changes in this PR. Should I move the hotfix to a separate PR?

0x4007 (approved):
Is the UUID being used? Otherwise everything looks fine @whilefoo rfc

scripts/github-action-permit-generator.ts
The-Gamer-01: Is this PR completed?and which ISSUE association code do I use?

0x4007:
> I understand this is a long running task which is why it was designed as an Action, but seems tricky to use in a real-case scenario.

I dont think this needs to be long running. So it makes sense to make an API endpoint hosted on Cloudflare Workers.

src/handlers/generate-erc20-permit.ts
```diff
@@ -54,6 +55,8 @@ export async function generateErc20PermitSignature(
       issueNodeId = contextOrPayload.payload.issue.node_id;
     } else if ("pull_request" in contextOrPayload.payload) {
       issueNodeId = contextOrPayload.payload.pull_request.node_id;
+    } else if (contextOrPayload.config.runId) {
```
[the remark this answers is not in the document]
hhio618: Thanks for the review, removed.

scripts/github-action-permit-generator.ts
```diff
@@ -0,0 +1,115 @@
...
+      amount: amount,
+      tokenAddress: env.EVM_TOKEN_ADDRESS,
+    }))
+  );
+
+  const config: PermitGenerationSettings = {
+    evmNetworkId: Number(env.EVM_NETWORK_ID),
+    evmPrivateEncrypted: env.EVM_PRIVATE_KEY,
+    permitRequests: permitRequests,
+    runId: runId,
```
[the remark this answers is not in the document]
hhio618: Removed.

src/types/env.ts
```diff
@@ -8,6 +8,10 @@ export const envSchema = T.Object({
   SUPABASE_KEY: T.String(),
   NFT_MINTER_PRIVATE_KEY: T.String(),
   NFT_CONTRACT_ADDRESS: T.String(),
+  EVM_NETWORK_ID: T.Optional(T.String()),
+  EVM_PRIVATE_KEY: T.Optional(T.String()),
+  EVM_TOKEN_ADDRESS: T.Optional(T.String()),
+  PAYMENT_REQUESTS: T.Optional(T.String()),
```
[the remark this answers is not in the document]
hhio618: Certainly, I've taken care of that in the following commits.

scripts/github-action-permit-generator.ts
```diff
@@ -0,0 +1,115 @@
...
+        console.error(message, ...optionalParams);
+      },
+    },
+    adapters: {} as ReturnType<typeof createAdapters>,
+  };
+
+  context.adapters = createAdapters(supabaseClient, context);
+
+  const permits = await generatePayoutPermit(context, config.permitRequests);
+  await returnDataToKernel(env.GITHUB_TOKEN, "todo_state", permits);
```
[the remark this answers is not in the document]
whilefoo: Because this script is only called manually and not by the kernel

0x4007:
What about the footer config code?

rndquu:
> What about the footer config code?

The actual comment is posted via https://github.com/ubiquity-os-marketplace/text-conversation-rewards, https://github.com/ubiquity-os/permit-generation is responsible solely for permits

hhio618:
> What about the footer config code?

Could you please provide more details or specifications for this task?

0x4007:
It's in the middle of the spec

PR #70, opened 2024-09-15
refactor: downgrade to ethers@v5

### LINKED ISSUE #60: Downgrade to `Ethers@v5`

cohow:
Resolves #60

I had to refactor a couple functions in `generate-erc20-permit.ts` due to issues with cognitive complexity, specifically I created three new functions `getPrivateKey` `getAdminWallet` and `getTokenDecimals`
I also had to include a yarn packageManager version in `package.json` because it wasn't letting me commit otherwise. let me know if you want the yarn version changed, I used whatever version corepack set.


package.json
```diff
@@ -121,5 +121,6 @@
       "import": "./dist/utils/index.js",
       "types": "./dist/utils/index.d.ts"
     }
-  }
+  },
+  "packageManager": "yarn@1.22.22+sha512.a6b2f7906b721bba3d67d4aff083df04dad64c399707841b7acf00f6b133b7ac24255f2652fa22ae3534329dc6180534e98d17432037ff6fd140556e2bb3137e"
```
0x4007, line 125: What if you used yarn 1.21? I believe I have this version and 1. I never use core pack 2. I never have to add this to package.json

0x4007 (commented):
Most of the changes generally look fine

package.json
cohow: yarn 1.21 works as well, I'm not sure why it did that to me but everytime i tried removing the `packageManager` line and committing it would throw an error about `Unstaged changes could not be restored due to a merge conflict!`.

0x4007: I never had this problem so I can't help you. 

Try reinitializing your yarn and remove core pack. Seems to introduce more problems than it fixes.

cohow: changing yarn versions to `1.12.1` seems to have fixed the problem! pushed a new commit to remove the packageManager line from `package.json`

Keyrxng (commented):
> Most of the changes generally look fine

I agree if anything was wrong with the downgrade the tests would have broken, can't really QA it much beyond that unless going the extra mile and testing it against the various overloads or using it to generate a permit (I don't think I've ever actually generated a QA permit before in any repo other than `pay.ubq.fi` and it uses it's own script, not this pkg)

cohow:
@0x4007 is this good to merge?

0x4007 (approved):
Let's test in production!

PR #72, opened 2024-10-06
Merge main into development

gentlementlegen:


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@ubiquibot/permit-generation",
-  "version": "1.3.1",
+  "version": "1.6.0",
```
0x4007, line 3: How is semver derived? Is it based on conventional commits prefixes?

gentlementlegen: Correct, it reads the commit names and deduces the version.

PR #73, opened 2024-10-07
fix!: changed rollup to tsup to simplify bundling

gentlementlegen:
Relates to https://github.com/ubiquity-os-marketplace/text-conversation-rewards/pull/144

### DIFF
--- a/tsup.config.ts
+++ b/tsup.config.ts
@@ -1,0 +1,11 @@
+import { defineConfig } from "tsup";
+
+export default defineConfig({
+  entry: ["src/index.ts", "src/core/index.ts", "src/utils/index.ts", "src/handlers/index.ts", "src/types/index.ts"],
+  splitting: false,
+  sourcemap: false,
+  clean: true,
+  dts: true,
+  target: "es2022",
+  format: ["cjs", "esm"],
+});
--- a/tsup.config.ts
+++ b/tsup.config.ts
@@ -3,7 +3,7 @@ import { defineConfig } from "tsup";
 export default defineConfig({
   entry: ["src/index.ts", "src/core/index.ts", "src/utils/index.ts", "src/handlers/index.ts", "src/types/index.ts"],
   splitting: false,
-  sourcemap: false,
+  sourcemap: true,
   clean: true,
   dts: true,
   target: "es2022",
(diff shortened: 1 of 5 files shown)


0x4007 (commented):
Pretty sure esbuild is superior what are you trying to accomplish here

gentlementlegen:
@0x4007 `tsup` is literally built on top of `esbuild`, we use it on most of out projects and it helps with bundling configuration made much simpler.

gentlementlegen:
https://github.com/egoist/tsup?tab=readme-ov-file#tsup

Needed to work within ESM projects as files were not compiled properly.

0x4007 (approved):

tsup.config.ts
```diff
@@ -0,0 +1,11 @@
+import { defineConfig } from "tsup";
+
+export default defineConfig({
+  entry: ["src/index.ts", "src/core/index.ts", "src/utils/index.ts", "src/handlers/index.ts", "src/types/index.ts"],
+  splitting: false,
```
0x4007, line 5: Tree shaking seems useful is that what this is

tsup.config.ts
```diff
@@ -0,0 +1,11 @@
...
+  sourcemap: false,
```
0x4007, line 6: This also seems very useful
```suggestion
  sourcemap: true,
```

0x4007:
You can merge after you make a decision on my review comment

0x4007:
You can merge after you make a decision on my review comment

tsup.config.ts
gentlementlegen: Not source that sourcemap are useful except for stack traces and debug which should not be included in production. It is useful if the code is minimized, which is not the case here.

tsup.config.ts
gentlementlegen: You mean `splitting` or `tree-shaking`?
Splitting is an option only relevant for `ESM`, not `CJS` and is not handled by `esbuild` which is why it's disabled. `treeShaking` is default on.
https://tsup.egoist.dev/#code-splitting
https://tsup.egoist.dev/#tree-shaking

tsup.config.ts
gentlementlegen: See also https://tsup.egoist.dev/#generate-sourcemap-file

0x4007: > Not sure that sourcemap are useful except for stack traces and debug which should not be included in production. It is useful if the code is minimized, which is not the case here.

Stack traces aren't magically removed from production I see them all the time from packages

PR #81, opened 2024-10-08
Feat: PR#69 adjustments

### LINKED ISSUE #80: Pull 69 Adjustments
> We can optimize and make an `ncc` build for faster cold start speed, as well as remove the "post comment to issue" action in another pull. 

_Originally posted by @0x4007 in https://github.com/ubiquity-os/permit-generation/pull/69#pullrequestreview-2353513136_

hhio618:
Resolves #80


scripts/github-action-permit-generator.ts
```diff
@@ -0,0 +1,116 @@
...
+
+/**
+ * Generates all the permits based on the current github workflow dispatch.
+ */
+export async function generatePermitsFromGithubWorkflowDispatch() {
+  const runId = github.context.runId;
+
+  // These are necessary to ensure the type checks and tests pass.
+  process.env["NFT_MINTER_PRIVATE_KEY"] = "";
+  process.env["NFT_CONTRACT_ADDRESS"] = "";
```
0x4007, line 21: Doesn't look right.

hhio618: Hey, I think this could be useful here if, down the road, someone needs to generate ERC721 tokens using this GitHub action. But for now, we can either stick with this approach or introduce a new environment schema, though that would require bigger changes to the codebase due to the type checks. Oh, and by the way, this actually relates to PR#69, not this one. I'll rebase once that PR is merged.

PR #96, opened 2024-10-30
Feat/workerize

### LINKED ISSUE #92: Convert plugin to cloudflare worker
Depends on https://github.com/ubiquity-os/permit-generation/issues/68

Related comments:
- https://github.com/ubiquity-os/permit-generation/pull/69#issuecomment-2399775638
- https://github.com/ubiquity-os/permit-generation/pull/69#issuecomment-2415718521

As far as I understand the reasons to convert this plugin from github action to cloudflare woker are:
1. It fits with the `pay.ubq.fi` rolled up feature because cloudflare functions located at `pay.ubq.fi` can call API endpoint and can not call (as well as get result) a github action
2. Other plugins (or kernel) will be able to call "generate permit" endpoint

So as a part of this issue we should:
- convert https://github.com/ubiquity-os/permit-generation to a cloudflare worker

Keyrxng:
Resolves #92


Keyrxng:
The bulk of the refactoring is done and tests are passing so it can be reviewed from here. One last change coming which is fixing encryption/decryption.

- I didn't solve the lib problem just hid it. So I need to swap `libsodium` or use one of the various versions (libsodium-wrappers-sumo, libsodium, -sumo, etc)
- Using a discriminated union (which requires the key `kind` within the `PermitRequest` schema; it has `type` already which is `ERC20`/`ERC721`. It's a worth while refactor for the benefit of dual schema validation)

Will do a complete overview tomorrow after the package problem is resolved

Keyrxng:
@UbiquityOS Summarize my pull request

Keyrxng:
`/ask` is not working as it should be.

1. Hallucinating against the spec because it never pulled the spec in.
2. The README inside the conversation block without distinction makes it appear as if it's part of the current conversation which should state `Pull Request` in the header but appears as `Issue`
3. It looks like it hasn't pull in the PR diff at all

Please see this [run](https://github.com/ubiquity-os-marketplace/command-ask/actions/runs/11605147726/job/32315045572)

0x4007:
> `/ask` is not working as it should be.
> 
> 1. Hallucinating against the spec because it never pulled the spec in.
> 2. The README inside the conversation block without distinction makes it appear as if it's part of the current conversation which should state `Pull Request` in the header but appears as `Issue`
> 3. It looks like it hasn't pull in the PR diff at all
> 
> Please see this [run](https://github.com/ubiquity-os-marketplace/command-ask/actions/runs/11605147726/job/32315045572)

@sshivaditya2019

Keyrxng:
> @Keyrxng, this task has been idle for a while. Please provide an update.

Just pushed code for 12 days straight, taking 2 days off (this is the 2nd)

Keyrxng:
> I looked through the logs and I don't understand how it created that spec based on real snippets of text from other unrelated places.

@0x4007 it's not real snippets of text in the logs. It's the readme which you wrote for this plugin aaages ago. It's literally just my comments and the readme that was pulled into the `formattedChat`. I'm unsure what embeddings search returned (unless you saw those SB logs or something) as they are not logged in that linked action run.

0x4007:
I searched some of the sentences and found results on our GitHub from various repos and various commenters.

Keyrxng:
> I searched some of the sentences and found results on our GitHub from various repos and various commenters.

Ah you mean pieces mentioned in the LLM response my bad I was meaning the formatted chat log, that goes back to us not seeing what embeddings search is actually returning, a quick fix is to log them independently so they can be assessed too. We should see the format chat, the embeddings results, the final ctx window and response I think.

Keyrxng:
Screenshot 2024-11-03 214135

Keyrxng:
> @Keyrxng, this task has been idle for a while. Please provide an update.

awaiting review

Keyrxng:
Can the reward be increased on this task to $1200 to match the issue that was [just closed](https://github.com/ubiquity-os/permit-generation/issues/71) that this actually resolves please? That would be great

Keyrxng:
@rndquu @gentlementlegen @zugdev @0x4007 

I'm assuming that this cannot be reviewed and merged until the rest of the infra has been built and ready to merge?

Tasks like https://github.com/ubiquity-os-marketplace/text-conversation-rewards/issues/64#issuecomment-2477262227 by the looks of it will be changing the current permit encoding schema so can review be started on this or is this on hold until other infra is ready?

gentlementlegen:
@Keyrxng you should assign people for review I forgot about this 😄 But I think it is testable. Eventually it will induce changes in plugins like `text-conversation-rewards` but can exist on its own, will test it.

gentlementlegen:
@Keyrxng I know there are already a lot of changes, but I think it should be worth adding the SDK here (which is basically the changes you made but manually).

src/types/plugin-inputs.ts
```diff
@@ -0,0 +1,58 @@
...
+    type: T.Literal("ERC20"), // should remove this ideally
+    userId: T.Number(),
+    amount: T.Number({ minimum: 1 }),
+    evmNetworkId: T.Number(),
+    tokenAddress: T.String(),
+    issueNodeId: T.String(),
+  }),
+  T.Object({
+    kind: T.Literal("ERC721"), // Discriminant key
+    type: T.Literal("ERC721"), // should remove this ideally
```
Keyrxng, line 28: @gentlementlegen what do you think about this, I'd remove `type` and keep `kind` as TypeBox needs `kind` and this PR will be part of a larger refactoring of all infra that uses this so I think this is acceptable.

Keyrxng:
> @Keyrxng you should assign people for review I forgot about this

Lmao it's still not possible to request review from anyone, think there is a settings issue or something idk.

> @Keyrxng I know there are already a lot of changes, but I think it should be worth adding the SDK here (which is basically the changes you made but manually).

Hmmmm idk because this needs to function as both a plugin and a standalone worker, is the SDK that extensible yet? I can try it though and we'll see I guess.

Keyrxng:
https://github.com/ubiquity-os/plugin-sdk/blob/004e1735a6c98d7bb015ac649bb2bbb57890f48f/src/server.ts#L32

The signature was my concern and I see we can bypass it so that's cool but not enough to support this worker I don't think.

https://github.com/ubiquity-os/plugin-sdk/blob/004e1735a6c98d7bb015ac649bb2bbb57890f48f/src/server.ts#L72

How it's bypassed would mean that we'd bypass all requests but we'd still like to enforce verification if it's a non-UI request I'd imagine. Which would mean I'd need to create two hono instances one to verify GH sided reqs and one to bypass UI reqs and implement some other handshake.

What do you think @gentlementlegen?

.dev.vars.example
gentlementlegen, line 1: Is it needed?

Keyrxng: I guess not no, I can fill it or remove it, lmk

gentlementlegen: Remove it if not needed, leave it with the required variables if needed.

Keyrxng:
> please resolve the conflicts? Also everything should be changed to use `bun` as a package manager for simplicity.

Conflicts resolved my friend. 

Can we create another task to migrate everything to `bun` please as this has been sat for a long time ready and you can make a clear spec on a bun migration for me as idk what that entails off the top of my head and I know it can also affect tests given what you've said on it before in regards to the kernel

Keyrxng:
What do you think @gentlementlegen, you okay with a new task for a bun migration or is it a requirement for this PR to move forward?

gentlementlegen:
It is fine to do it in another task although it should be quick I think.

rndquu (changes requested):
1. Somehow this request:
```json
{
    "settings": {
        "evmPrivateEncrypted": "INVALID",
        "permitRequests": [
            {
                "kind": "ERC721",
                "type": "ERC721",
                "userId": 119500907,
                "amount": 1,
                "evmNetworkId": 100,
                "tokenAddress": "0xe91D153E0b41518A2Ce8Dd3D7944Fa863463a97d",
                "issueNodeId": "0x123",
                "erc721Request": {
                    "contributionType": "issue.closed",
                    "keys": [
                        "GITHUB_ORGANIZATION_NAME", 
                        "GITHUB_REPOSITORY_NAME", 
                        "GITHUB_ISSUE_NODE_ID", 
                        "GITHUB_USERNAME", 
                        "GITHUB_CONTRIBUTION_TYPE"
                    ],
                    "values": ["test", "test", "123", "123", "contribution"],
                    "metadata": {
                        "GITHUB_CONTRIBUTION_TYPE": "contribution",
                        "GITHUB_ISSUE_NODE_ID": "123",
                        "GITHUB_ORGANIZATION_NAME": "test",
                        "GITHUB_REPOSITORY_NAME": "test",
                        "GITHUB_USERNAME": "123"
                    }
                }
            }
        ]
    }
}
```
returns this signed response:
```json
[
    {
        "tokenType": "ERC721",
        "tokenAddress": "0x0000000000000000000000000000000000000003",
        "beneficiary": "0x336C033842FA316d470e820c81b742e62A0765DC",
        "amount": 1,
        "nonce": "24409459659634628035242948860406449062914951070652615187495949183624535261750",
        "deadline": "115792089237316195423570985008687907853269984665640564039457584007913129639935",
        "signature": "0xb992feeb3fc8488a162d232fada403a55b4f0bf6150dba4bfea266b3850a27df62a61ef029cdd09c3d87016283a774f51de1e2fe235b970297bb45ebe05f8d071c",
        "owner": "0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266",
        "networkId": 100,
        "erc721Request": {
            "keys": [
                "0x3e4c8b8ab7e90cb1a6f2a4f809d6709c7eb62a413bdfb84f966832a075f4e4e4",
                "0x1f8205ede46f9b1cd2bb5d0f0d18e693848acc572c08bd312de5b83c0903cde0",
                "0x3aea1069a51bcd90299edb08a451733591bb49d89d05ab061f7933673466ccee",
                "0x6ead3148a7e9ddc486a0ca0542fa4a73f19558feda0b070d41a5c80573eeb9dc",
                "0x1f3d3cb9dfb4f8b84afce5bab9ed497e4967e37149f9045f96de46cfa7ad0c6f"
            ],
            "values": [
                "test",
                "test",
                "0x123",
                "123",
                "contribution"
            ],
            "metadata": {
                "GITHUB_ORGANIZATION_NAME": "test",
                "GITHUB_REPOSITORY_NAME": "test",
                "GITHUB_ISSUE_NODE_ID": "0x123",
                "GITHUB_USERNAME": "123",
                "GITHUB_CONTRIBUTION_TYPE": "contribution"
            }
        }
    }
]
```
although the `evmPrivateEncrypted` param is clearly invalid. The ERC20 permit generation throws the `Failed to decrypt a private key: Error: invalid input` error as expected in this case. Pls fix and add a unit test catching the `Failed to decrypt a private key` error on generating ERC721.

2. Pls fix failing knip CI

src/types/plugin-inputs.ts
```diff
@@ -0,0 +1,58 @@
...
+  eventPayload: WebhookEvent<T>["payload"];
+  settings: PermitGenerationSettings;
+  authToken: string;
+  ref: string;
+}
+
+const permitRequestSchema = T.Union([
+  T.Object({
+    kind: T.Literal("ERC20"), // Discriminant key
+    type: T.Literal("ERC20"), // should remove this ideally
```
rndquu, line 19: Aren't the `kind` and `type` properties the same? Can we remove the `kind` property? I can't find it in the latest `development` branch.

Keyrxng: Typebox needs `kind` so we should refer `type` if anything

rndquu: > Typebox needs `kind`

I don't understand. The `kind` property is missing in the latest `development` branch of https://github.com/ubiquity-os/permit-generation. Why is it appeared suddenly?

Keyrxng: I have added it in this PR to create a Typebox union so that we are handling the two different permits type more efficiently than what we were doing before. Now the permit inputs are being validated statically and at runtime.

Can simplify it back to what we had before if need be

rndquu: Possible values for `type` are `ERC20` and `ERC721`.

What are the possible values for the `kind` property?

Keyrxng: Exactly the same, they are the same prop with a different name. 

I was disappointed I couldn't assign `type` as the discrim key and had to use `kind`. 

Effectively they are identical in terms of value and application.

rndquu: @gentlementlegen help. I still can't understand why we need 2 exactly same properties here.

Keyrxng: we don't @rndquu.

I'm removing `type`

gentlementlegen: `kind` should not be needed technically, it is internal to typebox. I do not have the full context right now so I can't give you a proper answer but only one should be needed (and it is probably type).

Keyrxng: When building discriminated unions via typebox validators, it has the `kind` property as part of the schema by default and it is the discriminantKey. 

You can define a custom key as seen in the docs, but as standard `kind` is used.

https://github.com/jtlapp/typebox-validators?tab=readme-ov-file#discriminated-union-examples

For other union types, `kind` is not built-in the schema as it's identified some other way so for those `kind` would not be needed.

---

Simply put:

1. We use `type` as our custom key and remove `kind`.
2. We stick with the standard imposed by the package and remove `type`, relying on `kind` only.

Keyrxng:
> Somehow this request: ... returns this signed response:

@rndquu is it possible that you had `env.NFT_MINTER_PRIVATE_KEY` defined? It's likely if you just copy/pasted `.env.example`. Also, a check currently exists in `getErc721SignatureDetails` for the alt private key variable.

On that note, the two methods apparently require a different private key according to the old logic, is it still the case or should I refactor to use the same PK as the ERC20 method? I'm not totally clear on how NFTs are expected to operate so I kept this old logic since NFTs are not "rolled-out" yet.

Keyrxng:
With the `.env`/`.dev.vars` updated with a null value it does throw. Same with an incorrect hexlify value. I can repro but only with it defined in `.env` as `evmPrivateEncrypted` is not used by the erc721 method.

image

src/types/plugin-inputs.ts
Keyrxng: I've removed `kind` and assigned `type` as our custom key so `TypeBox Validators` should not complain with this approach according to the docs

0x4007:
Preliminary testing showed that azure functions are a lot more reliable than cloudflare workers. We migrated the kernel sometime within the last couple weeks and have discussed moving all plugins. @gentlementlegen rfc

The NFTs originally were envisioned to function as mementos and "achievements" similar like in gaming for working on repositories. Due to this, we would need a single shared Ubiquity owned key that can mint from the contract. 

However for the cryptocurrency side, the ERC20 key should be partner controlled because they are topping up the balance with their own money.

Keyrxng:
Understood, so things are good to remain as they are with the private keys.

No problem re: Azure, I've little experience there so perhaps it's better suited for someone else but I'll look over the spec for sure.

.dev.vars.example
```diff
@@ -0,0 +1,7 @@
+GITHUB_TOKEN= // this can be removed as it's not used anywhere besides actions
+SUPABASE_URL=
+SUPABASE_KEY=
+NFT_MINTER_PRIVATE_KEY=0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80
+NFT_CONTRACT_ADDRESS=0x0000000000000000000000000000000000000003
+X25519_PRIVATE_KEY=bHH4PDnwb2bsG9nmIu1KeIIX71twQHS-23wCPfKONls
+NODE_ENV=development
```
gentlementlegen, line 7: Probably `local` would make more sense?

.github/knip.ts
```diff
@@ -0,0 +1,14 @@
+import type { KnipConfig } from "knip";
+
+const config: KnipConfig = {
+  entry: ["build/index.ts"],
+  project: ["src/**/*.ts"],
+  ignore: ["src/types/config.ts", "**/__mocks__/**", "**/__fixtures__/**", "src/plugin.ts", "src/worker.ts", "src/helpers/validator.ts"],
+  ignoreExportsUsedInFile: true,
+  // eslint can also be safely ignored as per the docs: https://knip.dev/guides/handling-issues#eslint--jest
+  ignoreDependencies: ["@mswjs/data", "@typescript-eslint/eslint-plugin", "@typescript-eslint/parser", "ts-node", "tsx", "node"],
```
gentlementlegen, line 9: what is `tsx` used for

.github/workflows/formatting-checks.yml
```diff
@@ -0,0 +1,23 @@
...
+        uses: actions/checkout@v4
+
+      - name: Set up Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: "20.10.0"
+
+      - name: Run formatting checks
+        run: |
+          yarn install
```
gentlementlegen, line 22: Probably should use `bun`

gentlementlegen:
I think this pull-request is still useful to have the permit generation accessible from an endpoint?

If you can update workflows with up to date version adapted from https://github.com/ubiquity-os/plugin-template (remove bun, use our latest scripts etc)

Keyrxng:
ty @gentlementlegen

I updated the workflows and renamed the env var to `local`, should be good to go.

Keyrxng:
What or who do I need to do to make this move forward team?

0x4007:
@rndquu Can you handle the review

package.json
```diff
@@ -113,5 +124,6 @@
       "require": "./dist/handlers/index.mjs",
       "import": "./dist/handlers/index.js"
     }
-  }
+  },
+  "packageManager": "yarn@1.22.22"
```
gentlementlegen, line 128: Would be nice to use `bun` instead

src/worker.ts
```diff
@@ -0,0 +1,67 @@
...
+ *
+ * TweetNacl was written by the core authors of the original NaCl lib in C, and is a small lib
+ * that fits inside 100 tweets. It is closer to the metal than libsodium, and is more
+ * suitable for use in a worker environment.
+ *
+ * @ts-expect-error globalThis*/
+globalThis.importScripts = undefined;
+
+export default {
+  async fetch(request: Request, env: Env): Promise<Response> {
```
gentlementlegen, line 22: This looks very out of date compared to our plugin template can you confirm @Keyrxng or is there any reason to have this

Keyrxng: Yeah this can't really use the SDK because the intention is for it to be called from a UI, idk if you mean the structure of the fetch or not using the SDK exactly.

jest.config.json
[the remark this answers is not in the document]
Keyrxng: It was some time ago now mate, if it's going to hold back the PR I can update.

package.json
Keyrxng: Can we address this in another task dude?

src/worker.ts
gentlementlegen: Not sure what you mean by "be called from a UI", this is intended to be called by the FE in some cases indeed but also by plugins, or anything that needs a permit, I'll read the code more in detail but I don't see why the SDK can't handle all of this.

Keyrxng: It circles back around to routing although whilefoo addressed it [here more or less](https://github.com/ubiquity-os-marketplace/ubiquity-os-kernel-telegram/pull/48#discussion_r1959485981).

The implementation details from the FE perspective are pretty unclear right now anyway so yeah I'll use the SDK and we can deal with whatever pops up later.

Keyrxng: @gentlementlegen I'm guessing I should probably hardcode a return so that this endpoint can't actually be called right now right? Like what's the situation there, what events is it running on?

gentlementlegen: @Keyrxng It is not running on any event, it should be called via `fetch` (or at least that's what I understood) and return a permit.

Keyrxng: If we use the SDK to "call" this plugin via fetch wouldn't it need to match the same payload as the kernel expects for [`input-schema`](https://github.com/ubiquity-os/plugin-sdk/blob/development/src/types/input-schema.ts#L5) otherwise the SDK would error with [`Invalid body`](https://github.com/ubiquity-os/plugin-sdk/blob/5addd77fd7bb3949e39d80bca61a270a14a99c53/src/server.ts#L39)?

So would the kernel be forwarding requests from plugins or are plugins (as well as frontends) calling this directly using `fetch`, and if so, should `fetch` payloads have to match `input-schema`?

Or should they match the QA I've provided where it's far easier to `fetch` via the terminal instead of having to have `stateId, etc...`?

rfc too please @whilefoo @rndquu

Keyrxng: I did give it a whirl btw and hit that error which I realize requires _executive decisions_ lmao. 

If below is acceptable, then I'll happily push it.

```ts
export default {
  async fetch(request: Request, env: Env, executionCtx?: ExecutionContext) {
    return createPlugin<Context>(
      (context) => {
        const ctx = context as unknown as Context;
        const supabase = createClient(env.SUPABASE_URL, env.SUPABASE_KEY);
        ctx.adapters = createAdapters(supabase);
        const permits = generatePayoutPermits(ctx);
        return { permits };
      },
      manifest,
      {
        envSchema: envValidator.schema,
        postCommentOnError: false,
        settingsSchema: permitGenerationSettingsValidator.schema,
        logLevel: "debug",
        kernelPublicKey: env.KERNEL_PUBLIC_KEY,
        bypassSignatureVerification: process.env.NODE_ENV === "local"
      }
    ).fetch(request, env, executionCtx);
  }
}
```

whilefoo: This task should've been specced out better and should've been more thought out in advance as this should be clear from the start and not discussed in the PR. I imagined this wouldn't be a regular plugin because the frontend (pay.ubq.fi) can't send event payload, settings, signature...and they are also not needed for permit generation. 

It would just need encrypted funding wallet, network id and permit requests as we called them which include username, token type, token address and amount (basically this is what `text-conversation-rewards` passes to `permit-generation` in the config).

Keyrxng: That's what I've been thinking too so my current implementation should be good to go then.

A more accurate and considered spec can be made for how exactly both `plugins` and `frontends` will communicate with this worker.

I imagined that plugins would either:
A) chain this plugin within their config (kernel approved)
B) `fetch` while forwarding the kernel request that the originating plugin received (semi-kernel-approved)
C) This plugin would be the last plugin in _all_ chains and the kernel would tell it if it needs to create permits.

I imagined frontends would either:
A) Custom-event trigger a flow within the kernel which tells this worker to do it's thing--the permits are DB stored and polled from the UI
B) Direct `fetch` with the simplest of `settings` as it currently is:


```
curl -X POST http://localhost:4000 -H "Content-Type: application/json" -d '{"settings":{"evmPrivateEncrypted":"wOzNgt-yKT6oFlOVz5wrBLUSYxAbKGE9Co-yvT8f9lePsx7wJwPVugS9186zdhr1T4UpkpXvq9ii5M-nWfrydMnllSkowH4LirRZsHbvRVSvDoH_uh80p6HpwqDSG3g4Nwx5q0GD3H-ne4vwXMuwWAHd","permitRequests":[{"type":"ERC20","userId":106303466,"amount":1,"evmNetworkId":31337,"tokenAddress":"0xe91D153E0b41518A2Ce8Dd3D7944Fa863463a97d","issueNodeId":"0x123"}]}}'
```

package.json
Keyrxng: I followed [`plugin-template`](https://github.com/ubiquity-os/plugin-template/blob/development/package.json#L84) which omits it completely, I assume the reason why is because it forces you to enable `CorePack` if it's defined? 

Removed refs to yarn etc, using bun now dude.

Keyrxng:
QA:

Screenshot 2025-02-20 041351

Keyrxng:
Hey guys lmk what's needed to progress this, thank you.

src/handlers/erc721-permits/get-erc721-signature-details.ts
```diff
@@ -0,0 +1,98 @@
...
+  userId: number;
+  walletAddress: string;
+  nonce: string;
+  organizationName: string;
+  repositoryName: string;
+  username: string;
+  contributionType: string;
+}) {
+  if (!nftContractAddress) {
+    throw new Error(logger.error("NFT Address not found").logMessage.raw);
```
gentlementlegen, line 51: Is it necessary to wrap these in a `new Error`?

Keyrxng: you mean this instead? `throw logger.error("NFT Address not found")`?

src/handlers/erc721-permits/get-erc721-signature-details.ts
```diff
@@ -0,0 +1,98 @@
...
+  };
+
+  const domain = {
+    name: SIGNING_DOMAIN_NAME,
+    version: SIGNING_DOMAIN_VERSION,
+    verifyingContract: nftContractAddress,
+    chainId: evmNetworkId,
+  };
+
+  const signature = await adminWallet._signTypedData(domain, types, erc721SignatureData).catch((err: unknown) => {
```
gentlementlegen, line 92: You can omit `unknown`

.dev.vars.example
```diff
@@ -0,0 +1,7 @@
+GITHUB_TOKEN= // this can be removed as it's not used anywhere besides actions
```
gentlementlegen, line 1: Then shall it be removed?

package.json
```diff
@@ -26,46 +31,54 @@
...
     "@typescript-eslint/eslint-plugin": "^7.0.1",
     "@typescript-eslint/parser": "^7.0.1",
-    "cspell": "^8.4.0",
-    "eslint": "^8.56.0",
-    "eslint-plugin-sonarjs": "^0.24.0",
-    "husky": "^9.0.11",
+    "cspell": "8.14.4",
+    "eslint": "9.12.0",
+    "eslint-plugin-check-file": "2.8.0",
+    "eslint-plugin-sonarjs": "1.0.3",
```
gentlementlegen, line 67: We might want to use version 3

README.md
```diff
@@ -51,3 +51,14 @@ Return value should just be an array of generated permits. A separate module can
 [^2^]: We pass in the username for enhanced developer experience for plugin development. I believe it will be much easier to use GitHub usernames and let the system do the lookup.
 
 [^3^]: Mixed feelings on this lookup because it does not self encapsulate this module well. We already have an optional database write for the permit record, but this plugin WILL break if there is a database issue for reading the user's registered wallet, which can make testing and development more difficult here. The alternative is to pass in the user's wallet here, but then the kernel (or another plugin) needs to look up their wallet address (this seems like the better approach.)
+
```
gentlementlegen, line 54: Above, the `typescript` for `PermitGeneration` is wrong, add `=` after the type name

strings.json
```diff
@@ -0,0 +1,4 @@
+{
```
gentlementlegen, line 1: Not sure to understand what this is used for

wrangler.toml
```diff
@@ -1,8 +1,17 @@
-name = "ubiquibot-worker"
+name = "permit-generation"
```
gentlementlegen, line 1:
```suggestion
name = "ubiquity-os-permit-generation"
```

strings.json
Keyrxng: template scaffolding, removed.

.github/knip.ts
```diff
@@ -0,0 +1,14 @@
+import type { KnipConfig } from "knip";
+
+const config: KnipConfig = {
+  entry: ["build/index.ts", "src/worker.ts"],
+  project: ["src/**/*.ts"],
+  ignore: ["src/types/config.ts", "**/__mocks__/**", "**/__fixtures__/**", "src/plugin.ts", "src/worker.ts", "src/helpers/validator.ts"],
```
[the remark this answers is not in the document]
Keyrxng: `returnDataToKernel`, it was unclear whether or nut that was going to be needed but that was lazy I guess and so I've removed it

gentlementlegen:
@keyrxng Since you switched to bun could you upgrade to the latest version and change the `lockb` to `lock`

.github/knip.ts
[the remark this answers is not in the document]
gentlementlegen: I guess not since this won't be necessarily triggered by the kernel.

src/handlers/erc721-permits/get-erc721-signature-details.ts
gentlementlegen: For example yes, that's what we are usually doing in the plugins.

Keyrxng: throw using the logger and the tests do not pick it up as having thrown an error. I brought this up months ago during a logger PR while debugging the `the empty prop` bugs.

I made the changes and reverted them to show you but that has been an issue. The test suite will need beefed up to accommodate the new API/use-case of this worker. Can it be addressed as part of that PR task please?

Keyrxng:
> @Keyrxng Since you switched to bun could you upgrade to the latest version and change the `lockb` to `lock`

Done buddy ty for the review

whilefoo:
this is still backward-compatible, right? just the function signature changed (no more context, just permit requests), so we can keep using it as a library until everything is ready

Keyrxng:
> this is still backward-compatible, right? just the function signature changed (no more context, just permit requests), so we can keep using it as a library until everything is ready

It will still pass since the new params are still within `Context` and the 2nd arg is just unused but yeah it may have type-errored. I searched the repos for whats using the npm package and I think we are okay.

If we release as a new major version then even projects that are not tied to a particular version should not be affected right? I assumed that's what would be happening with things as you guys recently decided to version-lock dependencies.


https://github.com/ubiquity-os-marketplace/text-conversation-rewards/blob/0e7c9009fe3c421f2e4ddd4073c20e5a9c9b4682/src/parser/permit-generation-module.ts#L113

```ts
const permits = await generatePayoutPermit(
          {
            env,
            eventName,
            logger: permitLogger,
            payload,
            adapters: createAdapters(this._supabase, {
              env,
              eventName,
              octokit,
              config,
              logger: permitLogger,
              payload,
              adapters,
            }),
            octokit,
            config,
          },
          config.permitRequests
        );
```

1. 2nd arg is redundant, I'll add an arg but it will be unused.
2. Most of the first arg is also redundant, I've added `[key:string]:unknown` to stop any type errors in the invoking codebase.
3. `PermitRequests` previously sent `username` instead of `user.id`. We can either update the invoker or I can revert that change and fetch their GH ID?

image

Keyrxng:
As a matter of fact

- `env.NFT_CONTRACT_ADDRESS`
- `env.NFT_MINTER_PRIVATE_KEY`
- `env.X25519_PRIVATE_KEY`

are the only `env` vars required from `env` and if I'm not mistaken, this worker would have all of those already right? 

Partners just send us their `evmPrivateEncrypted` and we use our `X25519_PRIVATE_KEY` to decrypt? 

The NFT stuff, I'm sure this plugin would have as it's `env` vars too.

So I could really remove `env` too if we think about it?

---

The same goes for `adapters`, if payments/userbase etc are centralized to `UbiquityOS` Prod DB instance, then should we allow plugins to pass in their own DB?

The only method we use from it is `supabase.wallet.getWalletByUserId(userId)` so idk if it matters too much in terms of security but in terms of refactoring the method down to just `generatePayoutPermits(config.permitRequests)`

Keyrxng:
Further, shouldn't this plugin be responsible for pushing any permit requests it generates into the DB instead of invoking plugins like `text-conversation` doing it or not doing it potentially? 

I'm aware tasks are ongoing to implement pushing permits from that plugin as we speak but it doesn't make sense if more than that plugin will be requesting them.

Keyrxng:
Marked as off-topic because I was forward thinking things instead rather than for the npm package, my bad.

Keyrxng:
> PermitRequests previously sent username instead of user.id. We can either update the invoker or I can revert that change and fetch their GH ID?

This is the only important sentence to address before we can merge this and address the other problems in a separate PR

whilefoo:
> PermitRequests previously sent username instead of user.id. We can either update the invoker or I can revert that change and fetch their GH ID?

I think user id and wallet address should be passed in because we already fetch wallet address in `text-conversation-rewards` so it doesn't make sense to do it twice. `permit-generation` should be minimalistic and only focus on generating permits

Keyrxng:
I'm assuming you are talking about in regards to the npm package, but once permit generation is centralized to the worker I assume that it will be responsible for the user checks and pushing to the DB?

Kept `userId` and added `userWalletAddress` so we will need to update `text-conversation`.

I haven't touched the `encode/decode` methods or the types associated with them because that would break `pay.ubq` I'm sure so that can be left until the worker is rolled out.

Keyrxng:
@0x4007 @rndquu what do i need to do to get this merged dudes?

0x4007:
I think if it can pass @Whilefoo review it's fine.

Keyrxng:
Different when your salaried, you don't mind if these situations occur because your still earning at the end of the day. 

You don't mind waiting for shit to be clarified and the team to get on the same page with things. 

When your just a contributor it's a completely different story. And situations like this need to be avoided and remedied quickly where it cannot be avoided. 

PRs/tasks like these are what I've seen make newer devs stop contributing within the ecosystem more than anything else. 

---

If this feature isn't due to be used or rolled out or whatever then merge to a feature branch without a new release. 

Whatever it is that needs done, can it be done please.

0x4007:
@gentlementlegen Maybe you can bring it to the next step

src/handlers/generate-payout-permits.ts
```diff
@@ -0,0 +1,68 @@
...
+import { generateErc20Permit } from "./generate-erc20-permit";
+import { generateErc721Permit } from "./generate-erc721-permit";
+
+/**
+ * Generates a payout permit based on the provided context.
+ * @param context - The context object containing the configuration and payload.
+ * @param permitRequests
+ * @returns A Promise that resolves to the generated permit transaction data or an error message.
+ */
+// eslint-disable-next-line @typescript-eslint/naming-convention
```
gentlementlegen, line 13: Why is the rule disabled here?

src/handlers/generate-payout-permits.ts
```diff
@@ -0,0 +1,68 @@
...
+  const permits: PermitReward[] = [];
+  try {
+    const { permitRequests, evmPrivateEncrypted } = context.config;
+    logger.info(`Generating ${permitRequests.length} permits`);
+
+    for (const permitRequest of permitRequests) {
+      logger.info("Generating permit for: ", permitRequest);
+      const { amount, userId, nonce, evmNetworkId, tokenAddress, type: permitType, userWalletAddress } = permitRequest;
+      if (!userWalletAddress) {
+        throw new Error(logger.error(`No userWalletAddress provided for permit request: ${JSON.stringify(permitRequest)}`).logMessage.raw);
```
gentlementlegen, line 24: You can directly throw the `logger.error`

src/handlers/generate-payout-permits.ts
```diff
@@ -0,0 +1,68 @@
...
+      }
+    }
+
+    logger.info(`Generated ${permits.length} permits`);
+  } catch (error) {
+    logger.error(`Failed to generate permit: `, { er: String(error) });
+    throw error;
+  }
+
+  logger.info("Generated permits: ", { permits });
```
gentlementlegen, line 65: You could combine this log with the previous to avoid redundant logs (line 59)

src/helpers/validator.ts
```diff
@@ -0,0 +1,56 @@
...
+      path: "/evmPrivateEncrypted",
+      schema: permitGenerationSettingsSchema,
+      type: 0,
+      value: settings.evmPrivateEncrypted,
+    });
+  }
+
+  const permits = settings.permitRequests;
+
+  console.log("permits", permits);
```
gentlementlegen, line 30:
```suggestion
```

src/helpers/validator.ts
```diff
@@ -0,0 +1,56 @@
...
+
+  for (const permit of permits) {
+    if (!permitRequestValidator.test(permit)) {
+      for (const error of permitRequestValidator.errors(permit)) {
+        console.error(error);
```
gentlementlegen, line 35:
```suggestion
```

src/helpers/validator.ts
```diff
@@ -0,0 +1,56 @@
...
+  if (errors.length) {
+    throw { errors };
+  }
+
+  try {
+    const decodedSettings = Value.Decode(permitGenerationSettingsSchema, settings);
+    const decodedEnv = Value.Decode(envSchema, rawEnv || {});
+    return { decodedEnv, decodedSettings };
+  } catch (e) {
+    console.error("validateAndDecodeSchemas", e);
```
gentlementlegen, line 50:
```suggestion
```

wrangler.toml
```diff
@@ -1,8 +1,19 @@
-name = "ubiquibot-worker"
+name = "ubiquity-os-permit-generation"
 main = "src/worker.ts"
-compatibility_date = "2023-08-25"
+compatibility_date = "2024-09-25"
 node_compat = true
```
gentlementlegen, line 4: This is out of date: `compatibility_flags = [ "nodejs_compat" ]`

.github/workflows/knip-reporter.yml
```diff
@@ -6,22 +6,13 @@ on:
     types:
       - completed
 
+permissions: write-all
+
 jobs:
   knip-reporter:
     runs-on: ubuntu-latest
     if: ${{ github.event.workflow_run.conclusion != 'success' }}
     steps:
```
gentlementlegen, line 15: I think we can now replace `gitcoindev/knip-reporter@main` with `ubiquity/knip-reporter@main`

.github/workflows/update-configuration.yml
```diff
@@ -0,0 +1,21 @@
...
+  update:
+    name: "Update Configuration & Build"
+    runs-on: ubuntu-latest
+    permissions: write-all
+
+    steps:
+      - uses: ubiquity-os/action-deploy-plugin@main
+        with:
+          treatAsEsm: true
+          sourcemap: false
```
gentlementlegen, line 17:
```suggestion
          sourcemap: true
```

src/handlers/erc20-permits/get-erc20-signature-details.ts
```diff
@@ -0,0 +1,48 @@
...
+  nonce: string;
+  evmNetworkId: number;
+  evmPrivateEncrypted: string;
+  userId: number;
+  tokenAddress: string;
+  amount: number;
+  x25519privateKey: string;
+}) {
+  if (!userWalletAddress) {
+    throw new Error(logger.error("ERC20 Permit generation error: Wallet not found").logMessage.raw);
```
gentlementlegen, line 28: Can directly throw the logger error

src/handlers/erc20-permits/get-token-decimals.ts
```diff
@@ -0,0 +1,14 @@
...
+import { logger } from "../../helpers/logger";
+
+export async function getTokenDecimals(tokenAddress: string, provider: ethers.providers.Provider) {
+  try {
+    const erc20Abi = ["function decimals() public view returns (uint8)"];
+    const tokenContract = new ethers.Contract(tokenAddress, erc20Abi, provider);
+    return await tokenContract.decimals();
+  } catch (err) {
+    const errorMessage = `Failed to get token decimals for token: ${tokenAddress}, ${err}`;
+    logger.debug(errorMessage, { err });
```
gentlementlegen, line 11: You can make this a one-liner and throw the logger error (should probably not be `debug`)

src/handlers/erc721-permits/get-erc721-signature-details.ts
```diff
@@ -0,0 +1,96 @@
...
+  userId: number;
+  userWalletAddress: string;
+  nonce: string;
+  organizationName: string;
+  repositoryName: string;
+  username: string;
+  contributionType: string;
+}) {
+  if (!nftContractAddress) {
+    throw new Error(logger.error("NFT Address not found").logMessage.raw);
```
gentlementlegen, line 51: In this file it applies multiple time, same throw improvement can be made

.yarnrc.yml
```diff
@@ -1 +1 @@
-nodeLinker: node-modules
\ No newline at end of file
+nodeLinker: node-modules
```
gentlementlegen, line 1: I think this can be deleted since we are using bun now

src/handlers/erc20-permits/get-erc20-signature-details.ts
0x4007: Stop writing error in error messages. Also use warn instead of error unless it's something we'll need to triage later.

wrangler.toml
Keyrxng: It doesn't work with `compatibility_flags = [ "nodejs_compat" ]`
Screenshot 2025-03-12 091536


But does work with `node_compat = true`
image

src/handlers/erc20-permits/get-erc20-signature-details.ts
Keyrxng: https://github.com/ubiquity-os/permit-generation/pull/96#discussion_r1973068080
image

whilefoo:
> Final word on the `throw new Error` vs `throw logger` - I've highlighted why I've kept the `throw new Error` a couple of times during this PR.

I've usually not had this issue, example: https://github.com/ubiquity-os-marketplace/command-start-stop/blob/15a0fbb2cdde6831914ec1959784bf7d47cfa825/tests/main.test.ts#L236

wrangler.toml
whilefoo: that's weird, is your wrangler version up to date?

Keyrxng:
> > Final word on the `throw new Error` vs `throw logger` - I've highlighted why I've kept the `throw new Error` a couple of times during this PR.
> 
> I've usually not had this issue, example: https://github.com/ubiquity-os-marketplace/command-start-stop/blob/15a0fbb2cdde6831914ec1959784bf7d47cfa825/tests/main.test.ts#L236

I know I remember writing those tests (I just checked and some have reverted back to the old way around 2 months ago) and you explaining why `await expect(fn()).rejects` was better than `try{}catch{expect()}` and I'm certain I implemented those changes.

It stumped me for some time with this codebase and because everything is a grey area, and the fact, rndquu had mentioned about this repo before not having a specific test that would guarantee if anything was incorrect things would break. I have tried to not touch the tests for this reason.

Since it only impacts cosmetically how an error looks, I didn't want to mess with things for little gain in such an important codebase. The last time it had an error, permits were broken for weeks and the `rpc-handler` was hung, drawn and quartered for it. 

I see little gain from such an optimization when CF worker logs via the logger are hard to read anyway, but if I should throw logger, then may it be tasked out please or should I handle within this PR?

gentlementlegen:
Usually the `reject` works fine with custom errors, however what doesn't work is the `.toThrow(MyErrorClass)` due to TypeScript limitations when extending built-in classes (there are workarounds for this, like for example settings manually the prototype `Object.setPrototypeOf(this, MyError.prototype);`) which would make the tests work with custom errors.

wrangler.toml
Keyrxng: Bumping wrangler did indeed resolved it fellas

.github/workflows/jest-testing.yml
```diff
@@ -19,10 +19,32 @@ jobs:
...
         run: |
-          bun install --frozen-lockfile
-          bun run test
+          for i in {1..30}
+          do
+            if curl -s http://localhost:8545; then
+              break
+            fi
+            sleep 1
+          done || exit 1
```
0x4007, line 44: This looks wrong. 

Normally you would do an exponential backoff. Why does it fail in the first place?

Keyrxng: This isn't for error handling and I've yet to see the previous step fail on that note.

Anvil being ready to use varies from instance to instance depending on multiple factors, this is just a buffer to ensure it's ready for the suite. 

30 secs is enough time for the script to try multiple instances if need be.

Also this is taken from the other test suites in use with an Anvil module.

Keyrxng:
My bad @gentlementlegen for not spotting that earlier

Keyrxng:
Although @gentlementlegen, I fear that it'll interfere with the npm package because there are now technically two dist builds, one for the plugin and one for the package.

So when `publish-package` or `release-please` run it's going to push the worker too I'm sure.

I feel that it's a smart idea for this to be merged to a feature branch of it's own rather than to `development` and a plan of action put into place for actually migrating things. 

Like whether the `npm package` will be deprecated completely and putting into motion any changes in dependant plugins as they can hit the endpoint using the feature branch deployment.

gentlementlegen:
Wouldn't we want to stop using the package if that is deployed? Isn't it the final goal?

Keyrxng:
> Wouldn't we want to stop using the package if that is deployed? Isn't it the final goal?

I don't know what the final goal is or the steps to get there exactly. Which is why I'm suggesting a feature branch and a clear spec for next steps written by the team.

1. Is it the case, following the merge of this PR that effectively the `npm package` is deprecated? If so, then I didn't need to consider backwards compatibility and that logic can be removed. If not, then both are going to exist for a period of time.
2. If both exist, how is the infra that uses it migrating and when? Immediately after this is merged, otherwise things break?
3. If not going via a feature branch and ultimately affecting the current `npm package` build, then what if any, considerations need to be made now that the `dist` dir contains builds for two outputs. And should it be tasked separately.

whilefoo:
Yes, the final goal is to stop using it as a library and that `pay.ubq.fi` would call it as an API endpoint, but I think as an intermediate step we can make `text-conversation-rewards` call it directly.

whilefoo:
@0x4007 since https://github.com/ubiquity-os/permit-generation/issues/3 and https://github.com/ubiquity-os-marketplace/text-conversation-rewards/issues/64 are closed, reward rollup is not planned anymore? So what's the plan now?

Keyrxng:
> @0x4007 since #3 and [ubiquity-os-marketplace/text-conversation-rewards#64](https://github.com/ubiquity-os-marketplace/text-conversation-rewards/issues/64) are closed, reward rollup is not planned anymore? So what's the plan now?

lmao, wtf is going on with this DAO these days, does anyone really know what's going on? It's like the blind leading the blind honestly.

I hope that, like with other contributors recently, I still receive the reward if it's decided this is getting cancelled.

Disaster.

PR #97, opened 2024-10-31
fix: update octokit method calls to use the rest API

gentlementlegen:
Ensure that the `rest` calls of Octokit are used.


rndquu:
> Ensure that the `rest` calls of Octokit are used. @rndquu RFC, no logic has changed besides that.

1. `octokit.users.getByUsername()`
2. `octokit.rest.users.getByUsername()`

The difference between the lines above is that in the 1st case graphql API may be used?

gentlementlegen:
The `rest` is the modern way to use Octokit. The other way will be deprecated from what I've read. Our packages use the modern way (so does the SDK) so the octokit instance has to be used that way otherwise the methods don't exist. Octokit.graphql is also available indeed.

